PR #13, opened 2021-01-22
After installing kernel-ml in CentOS, update grub and reboot

Rumbles:
Fixes scylladb#12

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.idea
+.code
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-.idea
-.code
--- a/ansible-scylla-node/tasks/RedHat.yml
+++ b/ansible-scylla-node/tasks/RedHat.yml
@@ -14,7 +14,7 @@
   when: install_type == 'online' and (ansible_distribution == "CentOS" or ansible_distribution == "RedHat")
   become: true
 
-- name: Install ELRepo kernel
+- name: Install ELRepo kernel and reboot in to it if configured
   block:
     - name: install ELRepo key
       rpm_key:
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -0,0 +1,2 @@
+.idea
+.code
```
dyasny, line 2: No need for these

ansible-scylla-node/tasks/RedHat.yml
```diff
@@ -30,7 +30,37 @@
...
+        backup: yes
+        create: yes
+        regexp: '{{ item.regexp }}'
+        line: '{{ item.line }}'
+        state: present
+      with_items:
+        - regexp: 'GRUB_DEFAULT='
+          line: 'GRUB_DEFAULT=0'
+        - regexp: 'GRUB_TIMEOUT='
+          line: 'GRUB_TIMEOUT=30'
```
dyasny, line 49: Not sure such a long timeout is required. @vladzcloudius @tarzanek wdyt?

vladzcloudius: I agree. But why touching the timeout config in the first place?

dyasny:
@Rumbles Missing another step here - after the step installing the kernel, need to add a register: kernel_installed and only apply the grub changes and reboot when kernel_installed.changed. Otherwise the role will be rebooting nodes every time needlessly.

dyasny:
@Rumbles also, I don't see a reason to touch the grub timeout and make it so long. The sooner the nodes are back from reboot the better

.gitignore
Rumbles: I and other IDE users would disagree with you there

ansible-scylla-node/tasks/RedHat.yml
```diff
@@ -32,7 +32,38 @@
         name: kernel-ml
         state: present
         lock_timeout: 60
-  when: elrepo_kernel == True and install_type == 'online' and (ansible_distribution == "CentOS" or ansible_distribution == "RedHat")
+      register: kernel_installed
+
+    - name: Configure grub2 to boot first entry
```
tarzanek, line 37: this is needed only for Centos 7, for Centos 8 it's not - will it cause a problem in Centos 8 ?

tarzanek:
merging, I will retest Centos 8

PR #14, opened 2021-01-22
Allow user to provide SSL certs, wiki also needs updating

### LINKED ISSUE #11: When you add your own SSL certificates, nodes playbook ignores them and generates self signed ones
The [documentation](https://github.com/scylladb/scylla-ansible-roles/wiki/ansible-scylla-node:-Deploying-a-Scylla-cluster#ssl-configuration-for-scylla-nodes) states that:

>    To use your own CA, place it in ./ssl/ca/SCYLLA_CLUSTER_NAME-ca.pem and derived per-node certificates will be generated from it, if they are not present.
    To use your own node certificates, create a directory per hostname (same as in the inventory) in ./ssl/HOSTNAME and place the .pem file as HOSTNAME.pem and the .crt file as HOSTNAME.crt

When I tried this, my own certs were ignored and new ones generated. There currently isn't a check to see if certs were provided as the documentation describes (as far as I could see in ssl.yaml)

I have a fix I am currently testing, I will update if I think it is suitable

Rumbles:
Fixes #11 

In the wiki it mentions:

>    To use your own CA, place it in ./ssl/ca/SCYLLA_CLUSTER_NAME-ca.pem and derived per-node certificates will be generated from it, if they are not present.
>    To use your own node certificates, create a directory per hostname (same as in the inventory) in ./ssl/HOSTNAME and place the .pem file as HOSTNAME.pem and the .crt file as HOSTNAME.crt

But, as far as I could tell, neither of these things work.

I have added functionality for adding your own CA crt in `./ssl/ca` and pem/crt pairs for each host (as described in the wiki). But the wiki should be updated to reflect the fact that you need the CA crt, NOT the CA pem.

Further to this, there should be functionality to check if someone has added a CA pem, and then ansible should use that to generate pem/crt pairs for each host. This should be too hard to add, but I haven't the time right now and it's not my use case.


dyasny:
LGTM. Could you please open another issue about the missing functionality not covered by this PR, so it isn't lost?

Rumbles:
@dyasny I have added #15 to describe the missing own CA functionality I mentioned in my issue

dyasny:
@tarzanek @vladzcloudius it's up to you guys now :)

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -24,37 +37,70 @@
...
+        path: "./ssl/{{ item }}"
+      with_inventory_hostnames:
+        - scylla
+      register: hostname_folder_details
+      delegate_to: localhost
+
     - name: Create a directory for the key
       file:
-        path: "./ssl/{{ inventory_hostname }}"
+        path: "./ssl/{{ item }}"
```
vladzcloudius, line 56: Why this change was needed? What's "item"?

vladzcloudius (changes requested):

ansible-scylla-node/tasks/ssl.yml
Rumbles: On line 59 I have added:

```yaml
      with_items: "{{ hostname_folder_details.results }}"
      when: item.stat.isdir == False
```

This creates the item and means we can use that when creating the folder. It is required as this deals with the list of hostnames from the inventory. Maybe I'm doing this the wrong way? But I didn't see another way to check if all the folders existed and create them if they didn't...

vladzcloudius: Then, I wonder why we still need `inventory_hostname` in the namespace - it seems that you have eliminated all its usages but it still seems to be defined:
 
```
      with_inventory_hostnames:
        - scylla
```

@dyasny @tarzanek Please, comment.

Rumbles: it's defined by ansible, you cannot remove inventory_hostname...

Rumbles: Maybe I did it in the wrong way, I just found that this method worked in my case, but there might be a more graceful way to do this?

dyasny: `inventory_hostname` and `with_inventory_hostnames` are builtins, pointing to the hostnames ansible gets from the inventory. 

`inventory_hostname` is the current host that is being processed by ansible
`with_inventory_hostnames` is a generator for a list of these hostnames, so we can iterate over them. 

when iterating, the current processed list member is referred to as `item`

Rumbles: Actually thinking back to Friday this does explain why I saw ansible do something undesirable...

It is already be looping over the hosts, then I have created a second loop inside that loop as for 3 hosts, it printed out 9 lines in my terminal, sorry, I forgot I hadn't fixed that...

I'm not sure how to fix it right now, but it shouldn't loop in the loop like that...

tarzanek: I am confused, so how should it loop? 
what should be looped over what?
let's write it here in plain english and translate to ansible python after we all agree what it should do

Rumbles: At the moment, the ansible script loops over the hostnames from the inventory, but when I tried to detect whether the files/folders for each hostname existed and only create the files if they didn't exist I hit issues. I found this solution worked, but I had to add another loop inside, which probably isn't required, but it works, it just runs the operation more times than desired. If there is a way to check for the existence of the folders, the private key and the public part without having the additional loop created by lines 49/50 then it would be ideal.

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -1,20 +1,33 @@
 ---
 - name: generate a self-signed CA and use it for node certs
   block:
+    - name: Check CA folder exists
```
elcomtik, line 4: This is not necessary, as the ansible file module can do that by default in the next task.

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -1,20 +1,33 @@
...
         path: ./ssl/ca
         state: directory
+      when: ca_folder_details.stat.isdir == False
+
+    - name: Check CA pem exists
+      stat:
+        path: "./ssl/ca/{{ scylla_cluster_name }}-ca.crt"
+      register: ca_crt_details
 
     - name: Generate an OpenSSL private key for the CA.
```
elcomtik, line 20: Why generate a private key if we tested the presence of a certificate. The key may exist already.

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -1,20 +1,33 @@
...
       openssl_privatekey:
         path: "./ssl/ca/{{ scylla_cluster_name }}-ca.pem"
+      when: ca_crt_details.stat.exists == False
 
     - name: Generate an OpenSSL Certificate Signing Request for the CA
       openssl_csr:
         path: "./ssl/ca/{{ scylla_cluster_name }}-ca.csr"
         privatekey_path: "./ssl/ca/{{ scylla_cluster_name }}-ca.pem"
         common_name: "{{ scylla_cluster_name}}.internal"
+      when: ca_crt_details.stat.exists == False
```
elcomtik, line 30: Why generate a csr if we tested the presence of a certificate. The csr may exist already.

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -24,37 +37,70 @@
         issuer:
           commonName: "{{ scylla_cluster_name }}.internal"
         provider: selfsigned
+      when: ca_crt_details.stat.exists == False
```
elcomtik, line 40: Current module [docs](https://docs.ansible.com/ansible/latest/collections/community/crypto/x509_certificate_module.html) state that certificate is created if not present, this conditional is unnecessary.

ansible-scylla-node/tasks/ssl.yml
Rumbles: Hi sorry for the late reply, I think at the time I was having troubles with my particular setup, I use vault to generate my ca.crt, and machine crt/pem files and when I dropped those files in and ran this plan it overwrote the files I had setup

ansible-scylla-node/tasks/ssl.yml
Rumbles: IIRC The issue that it creates it when it is already present, but it shouldn't

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -1,20 +1,33 @@
...
+        path: ./ssl/ca
+      register: ca_folder_details
+
     - name: Create dir for the CA
       file:
         path: ./ssl/ca
         state: directory
+      when: ca_folder_details.stat.isdir == False
+
+    - name: Check CA pem exists
```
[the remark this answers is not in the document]
Rumbles: I just took the naming from the wiki [on this repo](https://github.com/scylladb/scylla-ansible-roles/wiki/ansible-scylla-node:-Deploying-a-Scylla-cluster#ssl-configuration-for-scylla-nodes), I do not mind about the naming I was just trying to get this to work :)

Rumbles:
As an update, I am aware this PR is behind master, I will try to rebase it soon.

I have just tested my deployment method again using master and I hit the same issues I hit when I first tested this, if you run the playbook after creating the `ssl/ca/CLUSTER_NAME-ca.crt` files and all the crt/pem files for the hosts (In my case I get these from my own vault, I cannot use self signed certs), the playbook overwrites my certificates with self signed certs and sets up the cluster using self signed certs (which is a dealbreaker for me unfortunately)

To clarify, this is how I get my ca and machine key/cert pairs:

```
resource "vault_pki_secret_backend_cert" "scylla_node" {
  count       = var.scylla_cluster_size
  backend     = "pki"
  common_name = aws_route53_record.scylla-nodes_dns[count.index].fqdn
  ip_sans     = [aws_network_interface.scylla_eni[count.index].private_ip]
  name        = "mesos"
  ttl         = "31557600" # 1 Year

  private_key_format = "pkcs8"
}

resource "local_file" "ca" {
  filename        = "${var.assets_dir}/ansible/ssl/ca/${local.cluster_name}-ca.crt"
  content         = vault_pki_secret_backend_cert.scylla_node[0].issuing_ca
  depends_on      = [null_resource.tidy_up_ssl_folders]
  file_permission = "0644"
}

resource "local_file" "public_cert" {
  count           = var.scylla_cluster_size
  filename        = "${var.assets_dir}/ansible/ssl/${aws_network_interface.scylla_eni[count.index].private_ip}/${aws_network_interface.scylla_eni[count.index].private_ip}.crt"
  content         = vault_pki_secret_backend_cert.scylla_node[count.index].certificate
  depends_on      = [null_resource.tidy_up_ssl_folders]
  file_permission = "0644"
}

resource "local_file" "private_key" {
  count             = var.scylla_cluster_size
  filename          = "${var.assets_dir}/ansible/ssl/${aws_network_interface.scylla_eni[count.index].private_ip}/${aws_network_interface.scylla_eni[count.index].private_ip}.pem"
  sensitive_content = vault_pki_secret_backend_cert.scylla_node[count.index].private_key
  depends_on        = [null_resource.tidy_up_ssl_folders]
  file_permission   = "0644"
}
```

@elcomtik with regards to the pem/crt naming you mentioned, I took this from the [wiki](https://github.com/scylladb/scylla-ansible-roles/wiki/ansible-scylla-node:-Deploying-a-Scylla-cluster#ssl-configuration-for-scylla-nodes), if this is the wrong naming convention, please let me know

I will try to sort out the loop issue I noticed and rebase to master soon

ansible-scylla-node/tasks/ssl.yml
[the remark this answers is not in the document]
Rumbles: Ah I see what you mean now! I have fixed on another branch, I will test it in the morning

Basically what I want to do is check if the ca.crt file already exists, if it does it shouldn't be generating the self signed pem...

ansible-scylla-node/tasks/ssl.yml
Rumbles: I removed this part as it wasn't needed, sorry :)

ansible-scylla-node/tasks/ssl.yml
Rumbles: You're right, I've removed this check

ansible-scylla-node/tasks/ssl.yml
Rumbles: That was my thought exactly, when I have ca.crt only it replaced it with self signed

tarzanek:
Thnx @Rumbles , so I see what the prob is:
https://docs.ansible.com/ansible/latest/collections/community/crypto/openssl_privatekey_module.html
says it will overwrite the existing files, if they don't match the defaults/settings of the module ...
so this is where your prob comes from and likely we should just use what is given, or check for the smallest common denominator of ssl certs that scylla supports ...
or just rely on the admin (and go with your approach) and if scylla will refuse the certificate, it will go back to admin to fix the certs in ansible folders

tarzanek:
I see options in the module such as 
format_mismatch 
or
regenerate 
to control this behaviour

PR #17, opened 2021-01-25
[#1927] Increased lock timeout for ech use of yum

### LINKED ISSUE #16: Yum commands often timeout if you're using a smaller server
During my PoC I was running the Scylla playbook on smaller nodes as I wanted to test the deployment process more than testing Scylla itself. I often hit issues when playbook steps failed because of yum lock being held and ansible timing out. This timeout can be increased fairly easily. PR to follow

Rumbles:
Fixes #16 

Doubles the lock timeout to avoid failing build steps when yum doesn't release it's lock within 30 seconds.

### DIFF
--- a/ansible-scylla-loader/tasks/RedHat.yml
+++ b/ansible-scylla-loader/tasks/RedHat.yml
@@ -9,6 +9,7 @@
   yum:
     name: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm"
     state: present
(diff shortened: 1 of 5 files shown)


dyasny (commented):
I've never run into this issue in my testing but the change is harmless, so LGTM

tarzanek:
lgtm, this basically doubles the time to wait for lock to be freed
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html#parameter-lock_timeout

PR #30, opened 2021-02-22
make auth optional

### LINKED ISSUE #19: settings for perftune for sq_split don't get applied
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/tasks/common.yml#L176
only prints the config, it won't set it

what should be done is (for 2019):

```
scylla_cpuset_setup --cpuset `/usr/lib/scylla/perftune.py --tune net --get-cpu-mask --nic eth0 --mode sq_split | /usr/lib/scylla/hex2list.py`
```

### LINKED ISSUE #29: add support for disabling authenthication

tarzanek:
fixes #29

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -24,6 +24,13 @@ reboot_after_kernel_upgrade: True
 # set this to false when the role isn't being applied to a full inventory reflecting the whole cluster
 full_inventory: true
 
+# set this to false to disable automatic
+# authenticator: PasswordAuthenticator
+# authorizer: CassandraAuthorizer
+# injected into scylla.yaml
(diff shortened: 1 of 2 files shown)


dyasny:
In general, this either is redundant, or needs to be very explicitly run AFTER `scylla_yaml_params` are applied, so that it overrides the defaults. Otherwise this will simply get overwritten by https://github.com/scylladb/scylla-ansible-roles/blob/7bc888f342c927cc30443da8c191f182a6ce8aa5/ansible-scylla-node/tasks/common.yml#L181

tarzanek:
yeah, the use case is only if you override scylla_yaml_params, will document this

dyasny:
also make sure to always cast to `bool` when checking a boolean variable, e.g.:

`{% if scylla_authentication is defined and scylla_authentication|bool %}`

PR #31, opened 2021-02-22
add cpuset

### LINKED ISSUE #19: settings for perftune for sq_split don't get applied

### LINKED ISSUE #28: cpuset is not run
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/tasks/common.yml#L166

just figures out the cpuset mask
but never sets it anywhere

so obviously for smaller machines sq_split won't be applied

tarzanek:
fixes #28
fixes #19

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -173,6 +173,15 @@
       perftune: "/usr/lib/scylla/perftune.py --tune net --get-cpu-mask --nic {{ scylla_nic }} --mode {{ perf_mode }} | /usr/lib/scylla/hex2list.py"
     when: is_reloc.stdout == 'non-reloc'
 
+  - name: set cpuset command
+    set_fact:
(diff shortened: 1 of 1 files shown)


dyasny:
LGTM

tarzanek:
can the "run perftune" block that happens afer cpuset be removed @dyasny ?

dyasny:
if all it does it print to stdout, then yes. We need an actual tuning here, not just a check

tarzanek:
ok, since I tested this I will commit and then look at perftune run - fwiw first scylla start runs perftune based on cpuset anyways

PR #32, opened 2021-02-22
Support ubuntu 20.04 (focal)

stevedrip:
Scylla server successfully starts on ubuntu focal after adding both this and the following PR:
https://github.com/scylladb/scylla-ansible-roles/pull/35

Not tested on production yet, nor have I attempted to install scylla manager.

### DIFF
--- a/ansible-scylla-manager/meta/main.yml
+++ b/ansible-scylla-manager/meta/main.yml
@@ -17,6 +17,7 @@ galaxy_info:
       - 7
   - name: Ubuntu
     versions:
+      - 20.04
       - 18.04
       - 16.04
   galaxy_tags: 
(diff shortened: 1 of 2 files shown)


dyasny:
Since this only changes meta, does it mean you've tested against Focal, and it worked?

stevedrip:
That's correct, though our team has only stood up a cluster and hasn't done any performance testing, tuning, or production deployments.

tarzanek:
per https://github.com/scylladb/scylla/issues/6343
support for 20.04 will be in 4.3.1

can we restrict this support only for certain versions? @dyasny ?

dyasny:
We can add a check at the start of the Debian.yml for scylla version and OS version. Something like:

```yaml
- fail:
    msg: "This Scylla version is incompatible with Ubuntu 20.x"
  when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version == "20" and scylla_edition == "oss" and scylla_actual_version >= "4.3.1"
```

All that is left is a quick `set_fact` for `scylla_actual_version` before this task, getting the installed version from apt or dpkg

tarzanek:
merging, since we support 20.04 now

PR #34, opened 2021-02-23
more verbosity to cleanup

tarzanek:


dyasny:
If you want those added `echo` commands to really work, you will need to `register` the result of the task into a variable and then print it out with `debug`:

```yaml
- name: my task
  shell: |
    echo TEXT!
  register: my_task_output

- debug:
    msg: "{{ my_task_output.stdout }}"
```

tarzanek:
correct, will fix

tarzanek:
alt. is to use -vv
but I guess we can just dump the output

Issue #2, opened 2020-09-30
perftune setup in scylla-node role


dyasny:
Need to run perftune.py

Up to 4 cores: mode=mq
5-8 cores: mode=sq
9+ cores: mode=sq_split

For reloc:
`sudo PATH=$PATH:/opt/scylladb/bin /opt/scylladb/scripts/perftune.py --tune net --get-cpu-mask --mode sq_split | /opt/scylladb/scripts/hex2list.py`
For non-reloc: ...

PR #35, opened 2021-02-25
Remove perftune covered by scylla_sysconfig_setup

stevedrip:
This block of code no longer works because the `hex2list.py` script no longer exists, as of 4.3.
https://github.com/scylladb/scylla/commit/85f76e80b43d8cfde4739a2a53c44c6e8850d793#diff-917434377d874a258a5aa772c8ec3b17ea3c966980e79f5fe70f387ee99dfdc1

I'm a bit unclear what was the motivation for adding this code block - perhaps scylla_cpuset_setup is or was unoptimal in the manner that it selected `perf_mode`?
https://github.com/scylladb/scylla-ansible-roles/issues/2

Without this block this seems to give a reasonable result for an Ec2 `i3.large` instance without the tuning:
```
cat /etc/scylla.d/cpuset.conf
# DO NO EDIT
# This file should be automatically configure by scylla_cpuset_setup
#
# CPUSET="--cpuset 0 --smp 1"
CPUSET="--cpuset 0-1 "
```

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -139,54 +139,6 @@
       scylla_sysconfig_setup --setup-nic-and-disks --nic {{ scylla_nic }}
   become: true
 
-- name: perftune related operations
-  block:
-  #TODO: additional testing on Debian/Ubuntu required.
-  - name: set perf_mode for 4 cores or less
-    set_fact:
-      perf_mode: mq
-    when: ansible_processor_vcpus <= 4
-
-  - name: set perf_mode for 5 to 8 cores
-    set_fact:
-      perf_mode: "sq"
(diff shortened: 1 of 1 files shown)


dyasny:
i3.large is too small for tuning to be meaningful IMO. You start seeing an effect after you go above 8 cores and more

tarzanek:
hmm, this however is needed for all other versions besides 4.3 ...
so I guess we need some versioning or some checks for scylla features here

tarzanek:
@stevedrip 
thanks for this, but we need cpuset setup to properly support bigger machines
I've fixed this in master, I will be closing this
let me know if any probs with the master fix

stevedrip:
That's great! Thanks for digging in.

PR #38, opened 2021-03-09
introduce back hex2list.py locally

tarzanek:


tarzanek:
this will fix 4.3 problem with cpuset

PR #39, opened 2021-03-10
Set correct values in io.conf

r4fek:
Without this change my io.conf file was empty after running example playbook. This way we ensure that `io_conf` value from `defaults/main.yml` will be used instead.


dyasny:
Why do you remove the check for existing io_properties?

dyasny:
The original idea was that if io_conf is already set on the nodes, there should be no need to re-set it by running the iotune benchmark again, when io_probe is enabled

PR #41, opened 2021-03-12
add example minorupgrade playbook

tarzanek:
fix rolling restart playbook task name


tarzanek:
I know it's simple and doesn't have rollback capability, but for minor upgrades it should be fairly safe

dyasny:
Great work there Lubos!

PR #48, opened 2021-04-23
example-playbooks: add restore_scylla_manager_backup

mmatczuk:
This patch adds a new playbook that let's you clone a cluster from a Scylla Manager backup.
It reqiures Scylla Manger 2.4 for the `scylla-manager-agent download-files` command.


mmatczuk:
I hand tested that in Scylla Cloud

```

**Comment by mmatczuk** [comment]
PTAL @tarzanek @vladzcloudius

**Comment by tarzanek** [comment]
LGTM, merging, respect and thank you @mmatczuk !

File: example-playbooks/restore_scylla_manager_backup/restore.yaml [hunk 1]
```diff
@@ -0,0 +1,135 @@
...
+
+    - name: Resore seeds list
+      lineinfile:
+        path: /etc/scylla/scylla.yaml
+        regexp: '^(\s+) - seeds:'
+        line: |
+          \g<1> - seeds: {{ _seeds.stdout }}
+        backrefs: yes
+
+    - name: Start Scylla service
```
tarzanek, line 132: fwiw I merged anyways
but ideally seeds should be started first
ala https://github.com/scylladb/scylla-ansible-roles/blob/master/example-playbooks/rolling_ops/rolling_restart.yml

we should make such OPS as whole cluster start available as some module command ... what do you think @sitano ?

dyasny: I saw that the playbook sets ALL nodes to seeds at some point when I skimmed through the new play.


PS/offtop - why not use the unirestore utility?

tarzanek: unirestore is not built for restoring with initial_token (so cloning from backup with restoring backup 1-1 , no need to cleanup or upload everything everywhere), though @vladzcloudius has a patch to get this support into unirestore too

also this depends heavily on next release of manager that supports dumping initial token

example-playbooks/restore_scylla_manager_backup/restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+      file:
+        path: "/var/lib/scylla/{{ item }}"
+        state: absent
+      loop:
+        - data
+        - commitlog
+        - hints
+        - view_hints
+
+    - name: Recreate Scylla directories
```
vladzcloudius, line 23: Is it really needed? AFAIK scylla is going to create them by itself which all corresponding permissions.
Better drop this.

example-playbooks/restore_scylla_manager_backup/restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+        line: |
+          \g<1> - seeds: {{ groups.all | join(',') }}
+        backrefs: yes
+
+    - name: Start Scylla service
+      service:
+        name: scylla-server
+        state: started
+
+    - name: Wait for CQL port
```
vladzcloudius, line 66: Very nice!

example-playbooks/restore_scylla_manager_backup/restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+- name: Restore schema
+  hosts: all[0]
+  become: true
+  gather_facts: false
+  tasks:
+    - name: Remove other seeds
+      lineinfile:
+        path: /etc/scylla/scylla.yaml
+        regexp: '^(\s+) - seeds:'
```
vladzcloudius, line 83: You want to set `node0` as a seeder to other nodes in the cluster too - not just for itself.

example-playbooks/restore_scylla_manager_backup/restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+      service:
+        name: scylla-server
+        state: stopped
+
+    - name: Download data
+      shell: |
+        scylla-manager-agent download-files -L {{ backup_location }} -n {{ host_id[inventory_hostname] }} -T {{ snapshot_tag }} -d /var/lib/scylla/data/ -K '*,!system_schema' --silent
+      become_user: scylla
+      async: 86400
+      poll: 5
```
vladzcloudius, line 122: What if it takes longer? This is not the best way to run an async task.
See here for a more robust way: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/async_status_module.html#async-status-module

example-playbooks/restore_scylla_manager_backup/restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+        port: 9042
+
+    - name: Stop Scylla service
+      service:
+        name: scylla-server
+        state: stopped
+
+    - name: Download data
+      shell: |
+        scylla-manager-agent download-files -L {{ backup_location }} -n {{ host_id[inventory_hostname] }} -T {{ snapshot_tag }} -d /var/lib/scylla/data/ -K '*,!system_schema' --silent
```
vladzcloudius, line 119: Is there any description/documentation what `scylla-manager-agent download-files` is going to do exactly?
What if copying fails? How many times it's going to retry? Is it going to validate checksums of copied files?

example-playbooks/restore_scylla_manager_backup/restore.yaml
vladzcloudius: @tarzanek Let's not rush merging things before all reviewers acked the PR. 
@dyasny It's not that unirestore couldn't be used here - it could. 
This playbook is a contribution from Michal - which is definitely welcome.
One nice thing it does is that it recreates the destination cluster which is beyond the unirestore scope. 

The ultimate plan is to re-write unirestore.py logic in Go and contribute it to SM.

In general it's not the best idea to mix deployment and copying things in the same PB - better have these as separate modules and then have a PB that calls these modules.

vladzcloudius (commented):
@mmatczuk I have a few comments.
It's a very nice PB. 
However, I have a few comments.

A few big comments:
1) Scylla role was supposed to be used instead of direct working with scylla-server service and tweaking of scylla.yaml.
2) The copying part is very unclear: robustness, retries, status check - I posted questions inline. In the long run merging unirestore.py code into SM looks like a more promising direction since it's going to allow us having both the backup and the restore functionality under the same roof (SM) without having additional ad-hoc layers like this PB.

example-playbooks/restore_scylla_manager_backup/restore.yaml
mmatczuk: I want to restore schema on single node, start it and then let the other node pull it.
I eventually restore original seed nodes.
I renamed it to `Temporarily set as the only seed`

example-playbooks/restore_scylla_manager_backup/restore.yaml
```diff
@@ -0,0 +1,135 @@
...
+      become_user: scylla
+      async: 86400
+      poll: 5
+
+    - name: Resore seeds list
```
tarzanek, line 124: Restore
typo

example-playbooks/restore_scylla_manager_backup/restore.yaml
mmatczuk: Yes it's configurable it uses the same logic as backup and is configured in the agent config file.

example-playbooks/restore_scylla_manager_backup/restore.yaml
mmatczuk: Unfortunately this does not work

```
Apr 27 12:25:54 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: /var/lib/scylla/data doesn't exist - skipping
Apr 27 12:25:54 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: /var/lib/scylla/commitlog doesn't exist - skipping
Apr 27 12:25:54 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: ERROR: 'disks' tuning was requested but no disks were found. Your system can't be tuned until the issue is fixed.
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: Traceback (most recent call last):
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: File "/opt/scylladb/scripts/libexec/scylla_prepare", line 67, in <module>
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: run("{} --options-file /etc/scylla.d/perftune.yaml".format(perftune_base_command()))
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: File "/opt/scylladb/scripts/scylla_util.py", line 329, in run
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: return subprocess.check_call(cmd, shell=shell, stdout=stdout, stderr=stderr, env=scylla_env)
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: File "/opt/scylladb/python3/lib64/python3.7/subprocess.py", line 363, in check_call
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: raise CalledProcessError(retcode, cmd)
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal scylla_prepare[15724]: subprocess.CalledProcessError: Command '['/opt/scylladb/scripts/perftune.py', '--tune', 'disks', '--dir', '/var/lib/scylla/data', '--dir', '/var/lib/scyll
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal systemd[1]: scylla-server.service: control process exited, code=exited status=1
Apr 27 12:25:55 ip-172-31-0-12.eu-central-1.compute.internal systemd[1]: Failed to start Scylla Server.
-
```

example-playbooks/restore_scylla_manager_backup/restore.yaml
mmatczuk: After reading [the docs](https://docs.ansible.com/ansible/2.9/user_guide/playbooks_async.html) it seems that what you suggest is intended to join on a task you spawned before 

> If you would like to perform a task asynchronously and check on it later you can perform a task similar to the following:

```
---
# Requires ansible 1.8+
- name: 'YUM - async task'
  yum:
    name: docker-io
    state: present
  async: 1000
  poll: 0
  register: yum_sleeper

- name: 'YUM - check on async task'
  async_status:
    jid: "{{ yum_sleeper.ansible_job_id }}"
  register: job_result
  until: job_result.finished
  retries: 30
```

Here we want to block on the download, I sure can change the t/o from 1 day to 1 week.

example-playbooks/restore_scylla_manager_backup/restore.yaml
vladzcloudius: Ugr... Right.  perftune.py runs before the initial invocation of scylla.

PR #51, opened 2021-05-04
Async extra

sitano:
The set of modules and plugins offers the mechanism to organize
named jobs on top of the Ansible standard async architecture. It
is fully compatible with Ansible native async wrapper and status.

async_task
===

    ---
    - hosts: localhost
      tasks:
        - name: Async recoverable job blah
          async_task:
            shell: |
              echo Hello &&
              sleep 10
            alias: blah
            async: 1000
            retries: 100
            delay: 5

instead of
===

    ---
    - hosts: localhost
      tasks:
        - name: Find if the task exists
          async_status_id: alias=blah
          register: job_result
        - name: Async job Blah
          shell: |
            echo Hello &&
            sleep 10
          async: 1000
          poll: 0
          register: async_task_obj
          when: not job_result.started
        - name: Register task alias
          async_alias: job=async_task_obj alias=blah
          when: not job_result.started
        - name: Wait for task to finish
          async_wait:
            alias: blah
            retries: 100
            delay: 5

instead of
===

    ---
    - hosts: localhost
      tasks:
        - name: Find if the task exists
          async_status_id: alias=blah
          register: job_result
        - name: Execute an async job Blah
          shell: |
            echo Hello &&
            sleep 10
          async: 1000
          poll: 0
          register: async_task_obj
          when: not job_result.started
        - name: Register task alias
          file:
            src: "{{ async_task_obj.results_file }}"
            dest: "{{ async_task_obj.results_file | dirname }}/jid_blah"
            state: link
          when: not job_result.started
        - name: Wait for task to finish
          async_status_id: alias=blah
          register: job_result
          until: job_result.finished
          retries: 100
          delay: 5
        - name: Cleanup
          async_status_id:
            alias: "blah"
            mode: "cleanup"
        - name: Print final object
          debug:
            msg: "{{job_result}}"

TODO
---
- check other command module modes
- test on remotes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.idea
+__pycache__
--- a/example-playbooks/async_extra/README.md
+++ b/example-playbooks/async_extra/README.md
@@ -112,6 +112,7 @@ Features:
 - always outputs started and finished fields
 - default `cleanup` mode cleanups only an alias if the job was specified
   by its name. if it was specified by a jid it deletes the whole state.
+- `full_cleanup` deletes job's state and all related links
 
 Quick start
 ---
--- a/example-playbooks/async_extra/README.md
+++ b/example-playbooks/async_extra/README.md
@@ -122,7 +122,7 @@ Simplest usage:
       async_status_id: alias="blah"
       register: job_result
 
-Example
+Example 1
 ---
 
 Possible implementation of the named recoverable jobs with this component alone:
@@ -158,7 +158,37 @@ Possible implementation of the named recoverable jobs with this component alone:
 This implementation do not handle overwrite of the aliases from other jobs
 at the step of registering a job alias. Neither it handles killed jobs.
 
-Example
+Example 2
+---
+
+The same as Example 1 but with `async_alias`:
+
+    - name: Find if the task exists
+      async_status_id: alias="blah"
+      register: job_result
+    - name: Execute an async job Blah
+      shell: |
+        echo Hello &&
+        sleep 60
+      async: 1000
+      poll: 0
+      register: async_task_obj
+      when: not job_result.started
+    - name: Register task alias
+      async_alias: job=async_task_obj alias=blah
+      when: not job_result.started
+    - name: Wait for task to finish
+      async_status_id: alias="blah"
+      register: job_result
+      until: job_result.finished
+      retries: 100
+      delay: 5
+    - name: Cleanup
+      async_status_id:
+        alias: "blah"
+        mode: "cleanup"
+
+Example 3
 ---
 
 An example shows an implementation of simple wait for an async job.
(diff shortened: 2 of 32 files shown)


sitano:
@vladzcloudius @tarzanek it's time to check this out

sitano:
Current issue that must be solved is the following:
- async_wrapped jid pid part means nothing, so that our status checker module have to do something with it. basically it needs to be able to ensure that  the wrapper is still alive and it has no refs to do that besides the jid string.
- if async task timedout based on async timeout (async_task_failure_timeout.yml) our status task can't detect wrapper already died due to ^^ so it will repeat all retries until it can
- running out of retries must be also be an error async_task_failure_retries.yml

sitano:
@vladzcloudius 

> @sitano There is another challenge here: you need to know if the task that has already finished and you wait on was successful or not. Do you have it covered already?

Yes, this is covered. The wrapper writes the finishing status to the job state file so we can see both the status, the outputs, and exit codes. The only thing that I want to be protected against if the wrapper suddenly dies. Basically, nothing bad would happen if the script died - the wrapper will handle it gracefully.

sitano:
All this is solved now.

> Current issue that must be solved is the following:
> 
>     * async_wrapped jid pid part means nothing, so that our status checker module have to do something with it. basically it needs to be able to ensure that  the wrapper is still alive and it has no refs to do that besides the jid string.
> 
>     * if async task timedout based on async timeout (async_task_failure_timeout.yml) our status task can't detect wrapper already died due to ^^ so it will repeat all retries until it can

If the wrapper died out of the sudden due to some unknown unexpected reason the jobs would fail to detect running wrapper and report an error about it. The async_status_id reports pid_missing == true. It can be cleaned up either by cleanup mode or manually by deleting jobs alias. 

>     * running out of retries must be also be an error async_task_failure_retries.yml

Fixed.

sitano:
The way it currently detects wrapper alive:
- kill -0 $pid, where `pid = jid.split(.)[1]`
- either way ps -elf | grep 'async_wrapper.py $jid'

vladzcloudius:
> All this is solved now.
> 
> > Current issue that must be solved is the following:
> > ```
> > * async_wrapped jid pid part means nothing, so that our status checker module have to do something with it. basically it needs to be able to ensure that  the wrapper is still alive and it has no refs to do that besides the jid string.
> > 
> > * if async task timedout based on async timeout (async_task_failure_timeout.yml) our status task can't detect wrapper already died due to ^^ so it will repeat all retries until it can
> > ```
> 
> If the wrapper died out of the sudden due to some unknown unexpected reason the jobs would fail to detect running wrapper and report an error about it. The async_status_id reports pid_missing == true. It can be cleaned up either by cleanup mode or manually by deleting jobs alias.
> 
> > ```
> > * running out of retries must be also be an error async_task_failure_retries.yml
> > ```
> 
> Fixed.

Got it. 
Thanks for the clarification. Now when I understand what has been done I think it exceeds my original expectations, @sitano .

I definitely agree that if the task hasn't reached to the successful completion for any reason, including getting killed, this should be interpreted as a failure.

However what we also want in this case is to have an APIs that would allow to "kill" and to "wait" for all already running tasks. The status of the "wait" should be statuses of corresponding tasks on all relevant nodes.

The status may be: 
 * Success
 * Failed
 * Killed
 * Unknown

sitano:
Kill is implemented. Wait was implemented alreay.

sitano:
See README.md.

tarzanek:
I did a quick review, I suggest to merge it and try to rewrite cleanup to use this and run a long cleanup OP, or interrupt the ssh and check resume of the cleanup job

sitano:
@tarzanek what I am doing on top of this work:
- adding support for `cmd` and `argv`
- adding support for `script` - requires support for a ansible.native.script
- rewriting cleanup play

PR #57, opened 2021-05-25
Molecule tests for ansible-scylla-node role

elcomtik:
- added ansible-lint config
- added .flake8 config
- added .gitignore config
- added .yamllint config
- added requirements[-dev].[in|txt] files for python packages
- added requirements.yml for dependent roles (currently none)
- added .github/workflows/default.yml with job for testing
  ansible-scylla-node role

This is an example of molecule tests for Scylla roles, it still needs a lot of work to function properly. The CI platform may be changed. Distros, ansible versions, and Scylla components need to be configured according to the support matrix.


elcomtik:
Ready for peer review. It still needs to fix a lot of ansible-lint issues and issue with perftune.py in docker.

PR #61, opened 2021-06-03
Async script

sitano:
- script support
- cmd support
- failed jobs warning
- run cleanup async

magic:

```
    - name: Run cleanup
      async_task:
        script:  files/cleanup.sh
        alias: scylla_cleanup
        async: 2592000
        retries: 43200
        delay: 60
      register: cleanup_output
```

Run
===

    $ ./run.sh localhost

or

    $ ANSIBLE_ACTION_PLUGINS=../async_extra/action_plugins ansible-playbook ./run_cleanup.yml -vvvv -l localhost -M ../async_extra/library

Note
===

it is critical to rework `./files/cleanup.sh` to exit with a non-zero error code if it actually fails. currently, it is not. but that's outside of the pr series.

### DIFF
--- a/example-playbooks/async_extra/action_plugins/async_task.py
+++ b/example-playbooks/async_extra/action_plugins/async_task.py
@@ -155,6 +155,9 @@ class ActionModule(ActionBase):
         running = self.find_running_task(alias, task_vars)
         if is_failed(running):
             result.update(running)
+            if is_finished(running):
+                result['warning'] = 'The job is finished and was not cleaned up properly. Please, delete job alias ' \
+                                    'manually to restart.'
             return result
 
         if running['started'] != 1:
@@ -189,6 +192,10 @@ class ActionModule(ActionBase):
                 delay=delay,
                 vars=task_vars
             ))
+
+            if is_failed(result):
+                result['warning'] = 'The job has failed and was not cleaned up properly. Please, delete job alias ' \
+                                    'manually to restart.'
         elif poll:
             raise AnsibleActionFail("poll is not supported yet")
 
@@ -388,3 +395,6 @@ class ActionModule(ActionBase):
 def is_failed(result):
     return 'failed' in result and result['failed']
 
+
+def is_finished(x):
+    return 'finished' in x and x['finished']
--- a/example-playbooks/async_extra/action_plugins/async_task.py
+++ b/example-playbooks/async_extra/action_plugins/async_task.py
@@ -94,6 +94,18 @@ started:
 
 class ActionModule(ActionBase):
 
+    _VALID_ARGS = frozenset((
+        # required
+        'alias', 'async',
+        # optional
+        'retries', 'delay', 'poll',
+        'cleanup',
+        # ansible.legacy.command
+        'shell', 'cmd', 'argv',
+        'chdir', 'executable', 'creates', 'removes', 'warn', 'stdin',
+        'stdin_add_newline', 'strip_empty_ends'
+    ))
+
     def run(self, tmp=None, task_vars=None):
         result = super(ActionModule, self).run(tmp, task_vars)
         del tmp  # tmp no longer has any effect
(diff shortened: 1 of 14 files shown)


sitano:
@tarzanek @vladzcloudius ^^

example-playbooks/async_extra/action_plugins/async_task.py
```diff
@@ -202,6 +273,56 @@ def run_async_task(self, vars):
...
+        #
+        # Once everything is encoded consistently, replace the script path on the remote
+        # system with the remainder of the command. This preserves quoting in parameters
+        # that would have been removed by shlex.split().
+        target_command = to_text(command).strip().replace(parts[0], tmp_src)
+
+        self._transfer_file(source, tmp_src)
+
+        # set file permissions, more permissive when the copy is done as a different user
+        self._fixup_perms2((self._connection._shell.tmpdir, tmp_src), execute=True)
```
vladzcloudius, line 313: What if this command fails (e.g. you don't have enough permissions). What will cleanup the state (e.g. delete the file transferred in the like 310 above)?

sitano: nothing. async jobs do not cleanup tmp folders.

example-playbooks/run_cleanup/action_plugins/async_alias.py
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
vladzcloudius, line 2: Why is this file under `run_cleanup`? This looks like something belonging to at least `async_extra`, no?

And in general why all these are under `example-playbooks`? A new module is not an "example". We should think of a better location. 

Something like `scylla-ansible-roles/modules`.

vladzcloudius (changes requested):

example-playbooks/run_cleanup/action_plugins/async_alias.py
sitano: > Why is this file under `run_cleanup`? This looks like something belonging to at least `async_extra`, no?

because `run_cleanup` is an independent playbook. there is a special folder structure where ansible looks for modules.

> And in general why all these are under `example-playbooks`? A new module is not an "example". We should think of a better location.

because all our ansible scripts sit in this `example-playbooks` directory. 

> Something like `scylla-ansible-roles/modules`.

yes good idea. we can rename it in a separate patch series.

example-playbooks/async_extra/action_plugins/async_task.py
vladzcloudius: Wouldn't it cause problems if one wants to re-run the command?

example-playbooks/run_cleanup/action_plugins/async_alias.py
vladzcloudius: > > Why is this file under `run_cleanup`? This looks like something belonging to at least `async_extra`, no?
> 
> because `run_cleanup` is an independent playbook. there is a special folder structure where ansible looks for modules. probably, I can replace those with soft-links to async-extra if yo uwill.
> 

Yes, that would be better. Let's keep all module related files under the same hierarchy and scripts that use it - under their own separate one.

 Also, I hope there is way to tell the Ansible where to look for extra modules. It would be ridiculous if there isn't.

> > And in general why all these are under `example-playbooks`? A new module is not an "example". We should think of a better location.
> 
> because all our ansible scripts sit in this `example-playbooks` directory.
> 
> > Something like `scylla-ansible-roles/modules`.
> 
> yes good idea. we can rename it in a separate patch series.

I see. Yes. Let's leave it for now here then.

example-playbooks/async_extra/action_plugins/async_task.py
sitano: no

example-playbooks/run_cleanup/action_plugins/async_alias.py
sitano: removed.
now this can be used: `ANSIBLE_ACTION_PLUGINS=../async_extra/action_plugins ansible-playbook ./run_cleanup.yml -vvvv -l localhost -M ../async_extra/library`.

vladzcloudius:
@tarzanek Could you, please, review. Let's see if can merge this?

tarzanek:
we can merge, I don't mind
I can test the cleanup change later this week once this is merged, but it looks OK and the only concern is with calling of cleanup playbook
so let's merge and I will adjust the callers/README later

PR #64, opened 2021-06-16
Add firewall control feature

ebenzecri:


ansible-scylla-node/tasks/common.yml
```diff
@@ -277,3 +277,20 @@
   become: true
   when: update_sysctl|bool is defined and update_sysctl|bool
 
+- name: Deactivate firewall
+  block:
+    - name: Deactivate front-end firewall services
+      service:
+        name: "{{ item }}"
+        state: stopped
```
tarzanek, line 285: what will happen if the machine will be restarted?
will the firewalls start back up?
what if ufw is not installed? (you have ignore errors, but I don't think this is a good solution, if you ignore error, then later you will figure out the firewall is not down and your connection doesn't work)

ansible-scylla-node/tasks/common.yml
```diff
@@ -277,3 +277,20 @@
   become: true
   when: update_sysctl|bool is defined and update_sysctl|bool
 
+- name: Deactivate firewall
```
tarzanek, line 280: I'd still be happy if you could test the whole deploy and ideally even with a 3 node cluster (I will try to merge the dev branch with molecule tests asap).
Why?
Because the task that checks if cluster is up is far far before this task and it would fail, since the firewall is up back then.
That said I think this task should be properly split to both RHEL and Debian yml files and do proper stuff there without ignore-ing the errors.
The this file will work as charm, since cluster will be created.

ebenzecri:
@tarzanek I've improved the code, please take a look. Thanks!

ansible-scylla-node/tasks/common.yml
```diff
@@ -277,3 +296,20 @@
...
+      loop:
+        - firewalld
+        - iptables_services
+        - ufw
+    - name: Flush all iptables rules
+      iptables:
+        flush: yes    
+  when: not firewall_enabled
+  ignore_errors: yes
+  become: true
```
tarzanek, line 315: the duplication of task is on purpose or forgotten from previous commit?

ansible-scylla-node/tasks/main.yml
```diff
@@ -1,5 +1,8 @@
 ---
 
+- name: Populate service facts
```
tarzanek, line 3: why is this needed?

tarzanek: ah, I see now, only if service is in facts, okay

tarzanek:
sorry for late review, so just clean up duplication and I think this can be merged
tia @ebenzecri !

ansible-scylla-node/tasks/common.yml
ebenzecri: @tarzanek Ignore this commit and use the last one. Sorry for the hassle.

ebenzecri:
@tarzanek Ignore this commit and use the last one which has all the corrections you pointed. Sorry for the hassle.

ebenzecri:
@tarzanek All done.

tarzanek:
thank you @ebenzecri :-) ! merging

PR #65, opened 2021-06-16
ELRepo related tasks improvements

ebenzecri:
RPM option used instead and some cosmetic changes applied.


tarzanek:
if it was tested, then lgtm, merging
thank you @ebenzecri !

PR #71, opened 2021-07-09
Version update

tarzanek:
centos 8 tests enabled


tarzanek:
added skip ntp and skip selinux options so I can enable centos 8 molecule tests
also added support for 3.8 monitoring

tarzanek:
fully tested with field cloud and latest versions

PR #72, opened 2021-07-12
Feature/upgrade

ebenzecri:
New main feature added: upgrade. **ansible-scylla-node** now will be able to upgrade a current Scylla installation. This is the first release.

### Upgrade process overview

1. Scylla current installation detection.
2. Validation of edition+version detected and also edition+version selected.
3. Version auto-correction if the upgrade can't be done in one simple step.
4. Pre-upgrade, upgrade, and post-upgrade tasks.
5. Node validation after the upgrade process is done

### Variables added and their default values
upgrade_version: false
upgrade_major: true
upgrade_rollback_on_failure: true
upgrade_cross_edition: false
upgrade_allow_user_manual_downgrade: false
upgrade_autocorrect_version: true
upgrade_generate_snapshots: false
upgrade_api_default_retries: 20
upgrade_api_default_delay: 15
upgrade_break_before_verification: 90
upgrade_backup_path: /root/scylla-backup

### TODO
1. Automatic rollback in case of failure.
2. Manual downgrade.
3. Cross-edition upgrade.
4. Iterative upgrade (useful when you need to do a major upgrade several times).


ebenzecri:
@tarzanek No rebase was needed :) Please, double-check to see if this is true.

tarzanek:
it created a merge commit, but it's ok

ansible-scylla-node/defaults/main.yml
```diff
@@ -12,6 +88,9 @@ scylla_nic: "{{ ansible_default_ipv4.interface }}"
 install_type: online
 
 # Defines if the node firewall should be enabled or not
+# Values:
+#  [default] false: all front-end firewall services will be disabled and iptables rules flushed
+#  true: keep node configuration as is
```
tarzanek, line 93: just a nitpick, commits should be self isolated and we shouldn't mix features / changes to them
so next time this change has to be a separate PR,  since it has nothing to do with feature/upgrade

ansible-scylla-node/handlers/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+
+- name: Clean yum metadata
+  command: yum clean metadata
+  args:
+    warn: no
+  become: true 
+
```
tarzanek, line 8: why?
what problem does this solve? yum should be able to update itself, resp. we can force update before installing

ansible-scylla-node/tasks/main.yml
```diff
@@ -1,16 +1,24 @@
...
 
+# Facts gathering
 - name: Populate service facts
   service_facts:
 
 - name: Populate package facts
   package_facts:
     manager: auto
 
+# Upgrade
```
tarzanek, line 11: so below will trigger update only when you flip the variable
then you will need to flip variable back so you will be able to be idempotent and not pickup new versions without knowing
OR is the upgrade version code idempotent?

ansible-scylla-node/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,305 @@
...
+# Scylla installation validation
+- name: Check existing Scylla installation operating system incompatibility
+  ansible.builtin.fail:
+    msg: "{{ ansible_facts['distribution'] }} {{ ansible_facts['distribution_version'] }} is not supported."  
+  when: >
+      ansible_facts['distribution'] |lower not in scylla_support or
+      scylla_support[ansible_facts['distribution'] |lower][ansible_facts['distribution_major_version']] is not defined or
+      ansible_facts['distribution_version'] is version(scylla_support[ansible_facts['distribution'] |lower][ansible_facts['distribution_major_version']]['minimum_version'],'<')
+
+- name: Check existing Scylla installation edition and version inccompatibility
```
tarzanek, line 115: typo

ansible-scylla-node/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,305 @@
+---
+
+# ---[ TODO ]-----------------------------------------------
+#    0) Test better Ubuntu/Debian distros for specific version upgrades. Check if apt module should be used or not.
+#    1) Downgrade support
+#    2) Cross-edition upgrade/downgrade
+# ---------------------------------------------------------
+
+# Scylla matrix for supported distros, editions and versions
```
tarzanek, line 9: this matrix likely needs a better place, since it can be shared with install too (e.g. refuse to install on old release)

ansible-scylla-node/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,305 @@
...
+    - upgrade_major
+    - upgrade_cross_edition
+
+
+# Upgrade
+- name: Trigger pre-upgrade tasks
+  include_tasks: pre_upgrade.yml
+
+- name: Trigger operating system specific upgrade tasks
+  include_tasks: "upgrade_{{ scylla_support[ansible_facts['distribution'] |lower][ansible_facts['distribution_major_version']]['template'] }}.yml"
```
tarzanek, line 269: I still cannot decide what I like better, above is likely lots of duplication and more maintainance probs
the other way would be single playbook per OS that would handle nitpicks/diffs for appropriate version

@vladzcloudius any take on above approach? (install takes the playbook per OS approach, this is what Eduardo suggested earlier, which gives us more possibilities, but will be harder to maintain and add new versions, so both approaches have their pros and cons)

ansible-scylla-node/tasks/upgrade/node_verification.yml
```diff
@@ -0,0 +1,65 @@
...
+- name: Check for joining nodes
+  ansible.builtin.uri:
+    url: "http://{{ scylla_api_address }}:{{ scylla_api_port }}/storage_service/nodes/joining"
+    method: GET
+  register: cluster_joining_nodes
+  until: cluster_joining_nodes.status == 200
+  retries: "{{ upgrade_api_default_retries |int }}"
+  delay: "{{ upgrade_api_default_delay |int }}"
+
+- name: Check for down nodes
```
tarzanek, line 31: this check will be tricky, since you will have nodes starting and you can only perform it AFTER all nodes are UN
(and in such case this test is moot)
install has a check for this

how big was a cluster that you tested this with? 3 nodes or more?

ansible-scylla-node/tasks/upgrade/main.yml
ebenzecri: @tarzanek The template is defined for each distro within the matrix, so two distros can share the same template file without any kind if duplicated content. Check Ubuntu, for example.

ansible-scylla-node/tasks/upgrade/pre_upgrade.yml
```diff
@@ -0,0 +1,165 @@
...
+  ansible.builtin.copy:
+    src: /etc/scylla/scylla.yaml
+    dest: "{{ upgrade_backup_path }}/{{ scylla_upgrade['id'] }}/etc/scylla.yaml"
+    mode: preserve
+    remote_src: yes
+  become: true
+
+# ----- TODO: Check if there are other files to backup and how -----
+
+# Stop Scylla service
```
tarzanek, line 160: just wanted to say kudoz for this .yml file
very nice fine grained stopping job :-)

ansible-scylla-node/tasks/upgrade/main.yml
ebenzecri: I agree, but so far 'upgrade' will be the only one to use it .

ansible-scylla-node/tasks/upgrade/upgrade_debian10.yml
```diff
@@ -0,0 +1,44 @@
+---
+
+# Add Scylla apt key
+- name: Add Scylla apt key
+  ansible.builtin.apt_key:
+    keyserver: hkp://keyserver.ubuntu.com:80
+    id: 5e08fbd8b5d6ec9c
```
tarzanek, line 7: we have a variable for that, haven't we?

ansible-scylla-node/tasks/upgrade/upgrade_debian9.yml
```diff
@@ -0,0 +1,52 @@
...
+    name: [apt-transport-https,dirmngr]
+    state: present
+  become: true
+
+
+# Add Scylla apt key
+- name: Add Scylla apt key
+  ansible.builtin.apt_key:
+    keyserver: hkp://keyserver.ubuntu.com:80
+    id: 5e08fbd8b5d6ec9c
```
tarzanek, line 15: same variable thing ... this id can change from time to time, hardcode is not good

ansible-scylla-node/tasks/upgrade/main.yml
ebenzecri: I ccccan't believe it. Fixing it.

ansible-scylla-node/tasks/upgrade/upgrade_redhat.yml
```diff
@@ -0,0 +1,49 @@
...
+    state: absent    
+  notify: Clean yum metadata  
+  become: true
+  when: upgrade_major
+
+
+# Create Scylla repositories file
+- name: Create Scylla repositories file
+  ansible.builtin.get_url:
+    url: "http://downloads.scylladb.com/rpm/centos/scylla-{{ scylla_upgrade['major_version'] }}.repo"
```
tarzanek, line 16: this url might also change, should be taken from some global variable

ansible-scylla-node/tasks/upgrade/upgrade_ubuntu16.yml
```diff
@@ -0,0 +1,52 @@
...
+  ansible.builtin.apt:
+    name: apt-transport-https
+    state: present
+  become: true
+
+
+# Add Scylla apt key
+- name: Add Scylla apt key
+  ansible.builtin.apt_key:
+    keyserver: hkp://keyserver.ubuntu.com:80
```
tarzanek, line 14: same stuff with keyservers and repo urls, should be taken from some global var

ansible-scylla-node/upgrade/main.yml
```diff
@@ -0,0 +1,305 @@
...
+
+# ---[ TODO ]-----------------------------------------------
+#    0) Test better Ubuntu/Debian distros for specific version upgrades. Check if apt module should be used or not.
+#    1) Downgrade support
+#    2) Cross-edition upgrade/downgrade
+# ---------------------------------------------------------
+
+# Scylla matrix for supported distros, editions and versions
+- name: Define Scylla support matrix
+  ansible.builtin.set_fact:
```
tarzanek, line 11: duplicated matrix? 
this asks for a single place where to take it from if possible

ansible-scylla-node/upgrade/main.yml
```diff
@@ -0,0 +1,305 @@
+---
```
tarzanek, line 1: is this whole file duplicated with tasks ?????
delete this????

ansible-scylla-node/tasks/upgrade/upgrade_debian10.yml
ebenzecri: This depends of the OS, using the same variable for the whole "Debian" family is a mistake.

tarzanek (changes requested):
is there some duplication between tasks/*
and just upgrade?

can you check above concern first @ebenzecri ?

ansible-scylla-node/tasks/upgrade/node_verification.yml
ebenzecri: The code copied our official upgrade doc. That's the way it's suggested to do it :/ I've tested it with a 3 nodes cluster

ansible-scylla-node/tasks/main.yml
tarzanek: so it can be overriden on the command line just for this run then? (so we can keep config files in git without adjusting their node_params.yml ? )
(try to think about the jenkins approach we use, how would it fit there and the git config store)

ebenzecri: You can use the command line to load the playbook instead. But, as far as I know, everything you define at CLI level has top precedence.

ansible-scylla-node/tasks/upgrade/upgrade_debian9.yml
ebenzecri: Then you change the OS template and you are done. There's no guarantee about having the same keyserver and key for Ubuntu and Debian. If you think about it, this only depends of the OS itself.

ansible-scylla-node/tasks/upgrade/upgrade_redhat.yml
ebenzecri: The repo URL is dynamically generated based on the selected version. The global variable is fixed string.

ansible-scylla-node/tasks/upgrade/upgrade_ubuntu16.yml
ebenzecri: The idea behind this is not to define this manually (since this is "encapsulated" per template), that's why there's no variable to override.

ansible-scylla-node/handlers/main.yml
ebenzecri: When removing a repository, beware that the metadata cache may still remain on disk until you run yum clean all. Use a notification handler for this.

_Source: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_repository_module.html_

The yum module does not support clearing yum cache in an idempotent way, so it was decided not to implement it, the only method is to use command and call the yum command directly, namely “command: yum clean all” https://github.com/ansible/ansible/pull/31450#issuecomment-352889579

_Source: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html_

ebenzecri: I don't really know if the handler is the best way to do it. In any case, we can try to find a better way to achieve this.

ansible-scylla-node/defaults/main.yml
elcomtik: In my opinion, this role should not interfere with the firewall at all. I would prefer if it wouldn't touch it so another role can manage it. There are many different firewall daemons used, and it shouldn't be adding complexity to this role. It can be handled like mdadm and swap is handled.

tarzanek:
please fix this:
```
diff --git a/ansible-scylla-node/tasks/upgrade/main.yml b/ansible-scylla-node/tasks/upgrade/main.yml
index de0bd90..fbc1519 100644
--- a/ansible-scylla-node/tasks/upgrade/main.yml
+++ b/ansible-scylla-node/tasks/upgrade/main.yml
@@ -176,7 +176,7 @@
     - name: Check if selected Scylla version is an upgrade or downgrade
       vars:
         detected_version: "{{ scylla_detected['major_version'] if upgrade_major else scylla_detected['version'] }}"
-        upgrade_version: "{{ scylla_upgrade['major_version'] if upgrade_major else scylla_upgrade['version'] }}"
+        scylla_upgrade_version: "{{ scylla_upgrade['major_version'] if upgrade_major else scylla_upgrade['version'] }}"
       ansible.builtin.set_fact:
         scylla_upgrade: {
           edition: "{{ scylla_upgrade['edition'] }}",
@@ -186,7 +186,7 @@
           id: "{{ scylla_upgrade['id'] }}",
           upgrade: false
         }
-      when: upgrade_version is version(detected_version,'<')
+      when: scylla_upgrade_version is version(detected_version,'<')

     - name: Check if downgrading is not allowed
       ansible.builtin.fail:
```

tarzanek:
also the upgrade task stops ALL nodes, while it should be sequential and ABORT if first node will have any sort of troubles, so we need to address this, I tested with 3 node cluster, now it stops all 3 nodes, which is obviously bad, upgrade needs to be rolling

tarzanek:
sample playbook for running the upgrade:
```
---

- hosts: scylla
  remote_user: scylla
  gather_facts: yes
  serial: 1
 
 
  vars:
    - upgrade_version: true
    - upgrade_major: true
    - scylla_edition: enterprise
    - scylla_version: latest
 
 
  roles:
    - ansible-scylla-node
```

ansible-scylla-node/tasks/upgrade/upgrade_redhat.yml
```diff
@@ -0,0 +1,49 @@
...
+    dest: /etc/yum.repos.d/scylla.repo
```
tarzanek, line 17: make sure this is in sync with what was used in nodes_params , resp. default install for role (see field cloud)

ansible-scylla-node/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,305 @@
+---
```
tarzanek, line 1: can we enforce serial: 1 directly here?
there is no way we will be running upgrade on multiple nodes at once

ebenzecri: AFAIK, you can't using inside a role since it's a playbook attribute. Here's the official doc: https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html

ansible-scylla-node/tasks/upgrade/upgrade_redhat.yml
ebenzecri: This was done on purpose since you don't need to set the repo: it will be set automatically based on your OS, Scylla edition and version. It will use the same destination file each time, avoiding more than one Scylla repo configured in the node.

If this mechanism is not used, you won't be able to use automation for upgrading since each major version has its own repo. This update is taking care about that so the user doesn't need to be worried about it.

IMO, same logic should be applied to the "install" feature (aka the original role).

tarzanek (approved):

tarzanek:
I will merge this, but we need some better sample for how to run this + make sure rolling restart is fool proof as in existing playbooks for minor upgrade
+ likely settle on common name for scylla upgrade repo file

PR #73, opened 2021-07-23
Set host for "Wait for CQL port"

simonfrey:
Explicitly set host for "Wait for CQL port" in restore_scylla_manager_backup/restore.yaml

Without this, the "Wait for CQL port" step timed out every time on gcs. (machine build from machine image)

### DIFF
--- a/example-playbooks/restore_scylla_manager_backup/restore.yaml
+++ b/example-playbooks/restore_scylla_manager_backup/restore.yaml
@@ -66,6 +66,7 @@
     - name: Wait for CQL port
       wait_for:
         port: 9042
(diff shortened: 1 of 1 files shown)


tarzanek:
@simonfrey thanks for this
what was the `ip add` output and output of `hostname` and output of /etc/hosts on the machine you had?

this should likely wait for the configured IP we configure for listen address, so we should take that variable

tarzanek:
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/defaults/main.yml#L99
so scylla_listen_address variable (in case someone overloads it)

tarzanek:
closing, this was moved to broadcast address variable

PR #76, opened 2021-08-04
Install correct version of openjdk on Debian

r4fek:
On Debian 10 (current stable) openjdk-8 isn't available, openjdk-11 should be installed instead.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -19,8 +19,8 @@ jobs:
         molecule_distro:
           #- centos7
           - centos8
-          #- debian9
-          #- debian10
+          - debian9
+          - debian10
           #- fedora
           #- ubuntu1604
           - ubuntu1804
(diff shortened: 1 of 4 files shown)


tarzanek:
I like this change, but it means our support matrix will cut out old Debian/Ubuntus
care to fix also
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/meta/main.yml#L15
and others?
(so we restrict the Debians and Ubuntus to supported ones? Also will this work on Debian 9? )

The support should ideally be at least for 9 and 10
same for Ubuntu, I guess 16.04 should get unsupported too (since it's standard support ended in April 2021)

tarzanek:
Per molecule Ubuntu 18, 20 work
care to enable molecule for Debian 9 and 10? (that would a proof your code just works (tm) )

tarzanek:
Sample how to enable molecule: https://github.com/scylladb/scylla-ansible-roles/commit/7c47452019fdb64d68c296448374a31c9bf59cfc

PR #77, opened 2021-08-04
Use correct docker repository

r4fek:
It works on both Debian and Ubuntu. Previously it worked only on Ubuntu.


tarzanek:
lgtm, merging, thnx @r4fek

PR #79, opened 2021-09-03
[FEAT] add support for custom data and log path in config template

toipa:
Having in mind previously configured raids it would be nice to be able to set the path for data and logs during the role application. 

This intends to do this with some variables that can be set to the user choice.

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -92,6 +92,13 @@ scylla_io_probe: True
 
 io_conf: 'SEASTAR_IO="--io-properties-file /etc/scylla.d/io_properties.yaml"'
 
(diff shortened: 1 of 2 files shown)


dyasny:
The way to set anything in scylla.yaml is already present, just add whatever options you want in `scylla_yaml_params` https://github.com/scylladb/scylla-ansible-roles/blob/facfe03c06531acc1006d2fdf352553019521f70/ansible-scylla-node/defaults/main.yml#L108

Implementing every one of the options in there separately would be pretty hard to maintain, IMO

micwiel:
@dyasny Adding `data_file_directories` to `scylla_yaml_params` doesn't work properly for me, what is proposed solution in this case?

PR #83, opened 2021-09-29
Old files and folder removal + new updates

ebenzecri:
The directory `upgrade` shout be removed since the valid one is under `tasks` directory. Also, I've set several improvements adding more files to the backup process.


tarzanek (approved):
lgtm thnx

PR #91, opened 2021-12-15
upgrade manager to 2.5 , various repo and apt cache fixes for Ubuntu

tarzanek:


ansible-scylla-node/defaults/main.yml
```diff
@@ -124,6 +124,7 @@ scylla_repo_keys:
   - 5e08fbd8b5d6ec9c
 
 scylla_manager_repo_keys:
+  - 5e08fbd8b5d6ec9c
```
vladzcloudius, line 127: Why keys are part of an Ansible code?
They should be an input parameter.

vladzcloudius (changes requested):

PR #107, opened 2022-04-01
ansible-scylla-node: bootstrap non-seeders serially

### LINKED ISSUE #106: ansible-scylla-node: start service serially
Bootstrapping Scylla nodes has to follow the documented procedure: https://docs.scylladb.com/operating-scylla/procedures/cluster-management/create-cluster/

In particular the next node should be not be started till the previous one is in UN state (we can wait till CQL port is listening) starting from seeders.

Today service on all nodes is started in parallel.

vladzcloudius:
Non-seeders have to be bootstrapped serially following
our standard procedure.

Fixes #106


vladzcloudius:
Tested on Field Cloud Training Facility

ansible-scylla-node/tasks/common.yml
```diff
@@ -217,21 +217,36 @@
...
+    - name: Get listen address
+      shell: |
+        grep "^\s*listen_address" /etc/scylla/scylla.yaml | cut -d":" -f2 | tr -d ' '
+      register: _listen_address
+
+    - name: Save listen_address as a fact
+      set_fact:
+        listen_address: "{{ _listen_address.stdout }}"
+
+    # Start all seeders together since we don't know which one is the "first" - so let them figure it by themselves
```
fee-mendes, line 229: I never understood this. We DO know which one is the first. It is rather simple. Quickly POC:

```yaml
- name: Test seed ordering
  gather_facts: false
  hosts: localhost
  vars:
    seed_list: []
    test:
      one:
        ipv4:
          address: "1.255.255.255"
      two:
        ipv4:
          address: "255.0.0.1"
      three:
        ipv4:
          address: "127.0.0.1"
  tasks:
    - name: print
      debug:
        msg: >
          "{{ test.one.ipv4.address }}"
          "{{ test.two.ipv4.address }}"
          "{{ test.three.ipv4.address }}"

    - name: addresses to seed array
      set_fact:
        seed_list: "{{ seed_list }} + [ '{{ item }}' ]"
      with_items:
        - "{{ test.one.ipv4.address }}"
        - "{{ test.two.ipv4.address }}"
        - "{{ test.three.ipv4.address }}"

    - name: print seed_list contents
      debug:
        var: seed_list

    - name: Print ordered IP addresses
      debug:
        msg: "{{ item }}"
      with_items:
        - "{{ seed_list | list | ipaddr('int') | sort | ipaddr }}"
```

fee-mendes (approved):
Approved, with remarks. :kiss:

ansible-scylla-node/tasks/common.yml
```diff
@@ -217,21 +217,36 @@
...
+    - name: Wait for CQL port on seeders
+      wait_for:
+        port: 9042
+        host: "{{ listen_address }}"
+      when: vars['ansible_'~scylla_nic].ipv4.address in scylla_seeds or inventory_hostname in scylla_seeds
+
+    - name: Start scylla non-seeds nodes serially
+      run_once: true
+      include_tasks: start_one_node.yml
+      loop: "{{ groups['scylla'] }}"
```
tarzanek, line 246: ok, now I see the trick :-)

ansible-scylla-node/tasks/common.yml
```diff
@@ -217,21 +217,36 @@
 
 - name: start_scylla_service dependent tasks
   block:
-    - name: start scylla seeds
+    - name: Get listen address
+      shell: |
+        grep "^\s*listen_address" /etc/scylla/scylla.yaml | cut -d":" -f2 | tr -d ' '
```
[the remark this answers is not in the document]
tarzanek: will that awk also trim whitespace as Vlads command does?

tarzanek (approved):
lgtm

tarzanek:
actually I have one concern
IF you run the role with multi fork , e.g. --forks 5 
won't there be a race that e.g. one of role installations didn't finish setup and from some other node the "run_once" code is called and starts scylla (that is not done configuring itself)?

I mean is there a hatch that ALL installed nodes will come to the critical section and ONLY then the run_once code starts?

(can be simulated by e.g. triggering some command on one of nodes to be slower by adding sleep to it)

ansible-scylla-node/tasks/common.yml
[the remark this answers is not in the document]
fee-mendes: Of course.

vladzcloudius:
> actually I have one concern IF you run the role with multi fork , e.g. --forks 5 won't there be a race that e.g. one of role installations didn't finish setup and from some other node the "run_once" code is called and starts scylla (that is not done configuring itself)?
> 
> I mean is there a hatch that ALL installed nodes will come to the critical section and ONLY then the run_once code starts?
> 
> (can be simulated by e.g. triggering some command on one of nodes to be slower by adding sleep to it)

I don't understand how this can be possible.
Ansible in its core doesn't allow that.

There is a hard stop between each task and 'fork' doesn't change that.

Therefore, unless I really don't understand how Ansible works, what you have described is not possible.

And if you are right and it's possible - meaning the next task can start before the previous task is finished on all (!) nodes, this whole role needs to be rewritten. In particular the way the service was started before this patch was totally broken.

fee-mendes:
Keep in mind Ansible default `fork` is 5 already, unless you are overriding it.

ansible-scylla-node/tasks/common.yml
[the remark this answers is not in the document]
vladzcloudius: Yeah. 
But I'm going to drop this whole thing since it turned out there is a `scylla_listen_address` already.

ansible-scylla-node/tasks/common.yml
vladzcloudius: It's not about their order but rather about inter-seeders dependency which can potentially be circular (which would be illegal but I don't want to care).

Let's see a few examples for 3 nodes cluster: N1, N2, N3

Ex 1:
N1_seeder: N2
N2_seeder: N2, N3
N3_seeder: N2

Here N2 is the "first" seeder and N3 needs to start after it and it's a seeder and N1 is a regular node.

Ex 2:
N1_seeder: N2
N2_seeder: N3
N3_seeder: N3

Here N3 is the "first" seeder and N2 is the "second" seeder.

Ex3:
N1_seeder: N1
N2_seeder: N1,N2
N3_seeder: N1,N2,N3

Here all 3 nodes are seeders and N1 has to be started the first.

Ex4:
N1_seeder: N2
N2_seeder: N3
N3_seeder: N1

Here all 3 nodes are seeders again but this is an illegal configuration.

The solution of the seeders bootstrap ordering is a matter of finding a root of a directed tree where vertices are seeders and there are edges V1->V2 if V2 seeds include V1.

And I don't want to start resolving this in Ansible.

So, IMO an easy workaround is to boot all seeders together and allow them figure this on their own since there are timeouts and retries and nodes are not allowed to join the existing cluster as seeder (I know Asias has it in the master and IMO in 2021.1 too - this means that there can be only a single seeder when you create a cluster unless you join with your own tokens). 

This means that the situation above can only happen during the initial cluster creation and therefore there will be a minimal streaming and a reasonable amount of seeds should be able to join even when started all together.

vladzcloudius:
I tested it with fork set to 10. But for 3 nodes it's quite irrelevant.

vladzcloudius:
@fee-mendes @tarzanek Please, review:

New in v2:
  - Use `{{ scylla_listen_address }}` instead of reading scylla.yaml.
  - Fixed a typo in start_one_node.yml "- name: Wait for CQL port ...": "of" -> "on".

fee-mendes (approved):
LGTM

ansible-scylla-node/tasks/start_one_node.yml
```diff
@@ -0,0 +1,15 @@
...
+    name: scylla-server
+    state: started
+  become: true
+  delegate_to: "{{ item }}"
+
+# Wait for at most 2 hours for a node to start - bootstrapping and the corresponding streaming can take quite long
+- name: Wait for CQL port on {{ hostvars[item]['listen_address'] }}
+  wait_for:
+    port: 9042
+    host: "{{ hostvars[item]['listen_address'] }}"
```
tarzanek, line 13: scylla_listen_address ?

tarzanek:
ok, this resolving of the variable is weird, but OK, yet another ansible weirdness

PR #112, opened 2022-04-06
ansible-scylla-node: resolve another composite variable

vladzcloudius:
Resolve vars['ansible_'~scylla_nic].ipv4.address into a fact
in order to be able to use it in a loop.


tarzanek:
lgtm, merging, thank you!

PR #113, opened 2022-04-06
ansible-scylla-node: resolve scylla_broadcast_address and use it

vladzcloudius:
When detecting if a node is a seed a 'broadcast_address' has
to be considered and not anything else.


tarzanek:
lgtm, merging

PR #114, opened 2022-04-07
always generate monitoring config and agent

### LINKED ISSUE #108: generate monitoring config and agent shouldn't depend on whether scylla service should be started
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/tasks/common.yml#L261
depends on starting scylla
but it shouldn't (so https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/tasks/common.yml#L269 should be moved above generation of monitoring and agent)

tarzanek:
fixes #108

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -276,16 +276,16 @@
       until: _result.status == 200
       retries: 90
       delay: 1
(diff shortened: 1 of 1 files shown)


tarzanek:
cool, after rebase the tests revealed I have a problem with my change ;-)
will fix

vladzcloudius (approved):

PR #115, opened 2022-04-07
Molecule fix

tarzanek:
this series fixes molecule and upgrades ansible for use in molecule to 5.6
it also enables Ubuntu 20 + enterprise and disables broken Centos 8 container (which lacks mirror access, so we need to find a different container to run for RHEL* versions)


tarzanek:
@vladzcloudius can we merge this? it will fix the molecule auto tests

PR #122, opened 2022-06-11
Add get scylla version playbook

vladzcloudius:
Scylla Role supports a 'latest' label as a Scylla Version (`scylla_version` parameter) however when we persist the Role state we should avoid storing `scylla_version: latest` in order to avoid unintentional SW upgrades next time we run a Role.

We should store the actual Scylla version instead.

This PR introduces a helper playbook that allows fetching a Scylla version and also "breaks" the default the way that would require to always explicitly provide a `scylla_version` parameter when a Role is executed.

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -152,7 +152,8 @@ scylla_dependencies:
 
 # Specify a Scylla version here (should be available in repo)
 # ex: scylla_version: 2019.1.9 or 3.2.1
-scylla_version: latest
+# Default is intentionally invalid version - it MUST be overridden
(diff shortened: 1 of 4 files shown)


vladzcloudius:
@tarzanek I couldn't figure if Molecule was broken by the PR (e.g. if the test doesn't explicitly specify a `scylla_version`) or it was broken before.
Let's take a look at this on Monday together.

tarzanek:
the tests reacted as expected, they broke, since 0.0.0 is version

tarzanek (changes requested):
please fix molecule as per my hints
(you can either pick a version or choose latest, both work for me)

PR #123, opened 2022-06-13
Only update apt cache, don't upgrade all packages for Ubuntu

tarzanek:
For monitoring and loader roles


tarzanek:
(similar as https://github.com/scylladb/scylla-ansible-roles/pull/119 but for other roles, too !)

PR #125, opened 2022-06-23
Debian: get versioning of scylla package working with role

### LINKED ISSUE #102: scylla node role: Ubuntu 20, if I select 2021.1.8 and in repo the latest is 2021.1.9 I can't get 2021.1.8 installed
Error seen:

```
10:12:36 fatal: [scylla-stage-us-central1-c-1]: FAILED! => {"cache_update_time": 1646125910, "cache_updated": false, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"       install 'scylla-enterprise=2021.1.8*'' failed: E: Unable to correct problems, you have held broken packages.\n", "rc": 100, "stderr": "E: Unable to correct problems, you have held broken packages.\n", "stderr_lines": ["E: Unable to correct problems, you have held broken packages."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nSome packages could not be installed. This may mean that you have\nrequested an impossible situation or if you are using the unstable\ndistribution that some required packages have not yet been created\nor been moved out of Incoming.\nThe following information may help to resolve the situation:\n\nThe following packages have unmet dependencies:\n scylla-enterprise : Depends: scylla-enterprise-server (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed\n                     Depends: scylla-enterprise-jmx (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed\n                     Depends: scylla-enterprise-tools (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed\n                     Depends: scylla-enterprise-tools-core (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed\n                     Depends: scylla-enterprise-kernel-conf (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Some packages could not be installed. This may mean that you have", "requested an impossible situation or if you are using the unstable", "distribution that some required packages have not yet been created", "or been moved out of Incoming.", "The following information may help to resolve the situation:", "", "The following packages have unmet dependencies:", " scylla-enterprise : Depends: scylla-enterprise-server (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed", "                     Depends: scylla-enterprise-jmx (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed", "                     Depends: scylla-enterprise-tools (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed", "                     Depends: scylla-enterprise-tools-core (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed", "                     Depends: scylla-enterprise-kernel-conf (= 2021.1.8-0.20220106.fcdf103fd-1) but 2021.1.9-0.20220208.86e3ea4df-1 is to be installed"]}
```

tarzanek:
fixes #102

### DIFF
--- a/ansible-scylla-node/tasks/Debian.yml
+++ b/ansible-scylla-node/tasks/Debian.yml
@@ -83,41 +83,24 @@
   - name: Figure out exact Scylla version
     block:
     - name: Get versions of {{ scylla_package_prefix }} package
-      shell: apt list -a {{ scylla_package_prefix }} | grep {{ scylla_version }}
+      shell: apt list -a {{ scylla_package_prefix }} | egrep "{{ scylla_version }}[\-]?"
       register: aptversions
 
     - name: Search for version of {{ scylla_package_prefix }} package
       set_fact:
         tempsversion: "{{ aptversions.stdout | regex_findall(regexp) }}"
       vars:
-          regexp: ' ([0-9\.]+)-([a-f0-9\.]+)-([0-9]) '
+        regexp: ' ([0-9\.]+)-([a-f0-9\.]+)-([0-9]) '
 
     - name: Set {{ scylla_package_prefix }} version if only one version found
       set_fact:
         sversion: "{{ tempsversion }}"
       when: tempsversion | length == 1
 
-    - name: Requery for {{ scylla_package_prefix }} version if multiple versions found
-      block:
-        - name: query for version with - at the end of {{ scylla_version }}
-          shell: apt list -a {{ scylla_package_prefix }} | grep {{ scylla_version }}-
-          register: aptversions
+    - name: Error out, wrong Scylla version was passed(more than a single or none version matches), please fix it!"
+      ansible.builtin.command: /bin/false
+      when: tempsversion | length != 1
 
-        - name: Search for version of {{ scylla_package_prefix }} package with - postfix
-          set_fact:
-            rtempsversion: "{{ aptversions.stdout | regex_findall(regexp) }}"
-          vars:
-              regexp: ' ([0-9\.]+)-([a-f0-9\.]+)-([0-9]) '
-
-        - name: Set scylla version if only one version with - found
-          set_fact:
-            sversion: "{{ rtempsversion }}"
-          when: rtempsversion | length == 1
-
-        - name: Error out, wrong Scylla version was passed, please fix it!
-          ansible.builtin.command: /bin/false
-          when: rtempsversion | length > 1
-      when: tempsversion | length > 1
     - name: Set Scylla Full version
       set_fact:
         scylla_full_version: "{{ sversion }}"
@@ -136,8 +119,7 @@
       apt:
         name: "{{ scylla_package_prefix }}={{ scylla_version }}*" # use {{ scylla_full_version[0][0] }}-{{ scylla_full_version[0][1] }}-{{ scylla_full_version[0][2] }} here ?
         state: present
-        allow_downgrade: yes # this is questionable, do we want to allow the way back?
-# remove pin?
+        allow_downgrade: yes
     when: scylla_version != 'latest'
 
   become: true
--- a/ansible-scylla-node/tasks/Debian.yml
+++ b/ansible-scylla-node/tasks/Debian.yml
@@ -70,17 +70,76 @@
       state: latest
     when: scylla_version == 'latest' and scylla_edition == 'enterprise'
 
-  - name: Install specified OSS Scylla
-    apt:
-      name: "scylla={{ scylla_version }}*"
-      state: present
+  - name: Set Scylla {{ scylla_edition }} prefix
+    set_fact:
+      scylla_package_prefix: "scylla"
     when: scylla_version != 'latest' and scylla_edition == 'oss'
 
-  - name: Install specified Enterprise Scylla
-    apt:
-      name: "scylla-enterprise={{ scylla_version }}*"
-      state: present
+  - name: Set Scylla {{ scylla_edition }} prefix
+    set_fact:
+      scylla_package_prefix: "scylla-enterprise"
     when: scylla_version != 'latest' and scylla_edition == 'enterprise'
+
+  - name: Figure out exact Scylla version
+    block:
+    - name: Get versions of {{ scylla_package_prefix }} package
+      shell: apt list -a {{ scylla_package_prefix }} | grep {{ scylla_version }}
+      register: aptversions
+
+    - name: Search for version of {{ scylla_package_prefix }} package
+      set_fact:
+        tempsversion: "{{ aptversions.stdout | regex_findall(regexp) }}"
+      vars:
+          regexp: ' ([0-9\.]+)-([a-f0-9\.]+)-([0-9]) '
+
+    - name: Set {{ scylla_package_prefix }} version if only one version found
+      set_fact:
+        sversion: "{{ tempsversion }}"
+      when: tempsversion | length == 1
+
+    - name: Requery for {{ scylla_package_prefix }} version if multiple versions found
+      block:
+        - name: query for version with - at the end of {{ scylla_version }}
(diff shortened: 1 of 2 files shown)


tarzanek:
tests done were basic downgrade and upgrade
( 2021.1.11 to 2021.1.9 and 2021.1.9 to 2021.1.10 )

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -70,17 +70,76 @@
...
+
+  - name: Figure out exact Scylla version
+    block:
+    - name: Get versions of {{ scylla_package_prefix }} package
+      shell: apt list -a {{ scylla_package_prefix }} | grep {{ scylla_version }}
+      register: aptversions
+
+    - name: Search for version of {{ scylla_package_prefix }} package
+      set_fact:
+        tempsversion: "{{ aptversions.stdout | regex_findall(regexp) }}"
```
vladzcloudius, line 91: weird identation

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -70,17 +70,76 @@
...
+      vars:
+          regexp: ' ([0-9\.]+)-([a-f0-9\.]+)-([0-9]) '
+
+    - name: Set {{ scylla_package_prefix }} version if only one version found
+      set_fact:
+        sversion: "{{ tempsversion }}"
```
vladzcloudius, line 97: And here

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -70,17 +70,76 @@
...
+      when: tempsversion | length == 1
+
+    - name: Requery for {{ scylla_package_prefix }} version if multiple versions found
+      block:
+        - name: query for version with - at the end of {{ scylla_version }}
+          shell: apt list -a {{ scylla_package_prefix }} | grep {{ scylla_version }}-
```
vladzcloudius, line 103: I guess you are trying to handle the case where you are either given the full scylla version like `2021.1.11-0.20220516.8e6c9917c` or a short version like `2021.1.11`, right?

I wonder why these two case can't be folded into a single regular expression?

Something like `egrep "{{ scylla_version }}[\-]?"`

image

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -70,17 +70,76 @@
...
+        src: templates/apt-pin-file.j2
+        dest: "/etc/apt/preferences.d/99{{ scylla_package_prefix }}"
+        owner: root
+        group: root
+        mode: '0644'
+    - name: Install Scylla
+      apt:
+        name: "{{ scylla_package_prefix }}={{ scylla_version }}*" # use {{ scylla_full_version[0][0] }}-{{ scylla_full_version[0][1] }}-{{ scylla_full_version[0][2] }} here ?
+        state: present
+        allow_downgrade: yes # this is questionable, do we want to allow the way back?
```
vladzcloudius, line 139: > do we want to allow the way back?

This depends and this needs to be resolved on an `upgrade` task level - not here.

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -70,17 +70,76 @@
...
+# remove pin?
```
vladzcloudius, line 140: What would happen if you don't remove a pin and install a different version (upgrade use case)? Have you tested this?

vladzcloudius (changes requested):

ansible-scylla-node/tasks/Debian.yml
tarzanek: so in such case I should allow it, removing this comment

ansible-scylla-node/tasks/Debian.yml
tarzanek: removing this comment

ansible-scylla-node/tasks/Debian.yml
tarzanek: reason was different, it was to error out if wrong version was provided (one that doesn't have a "-" at the end and doesn't match anything)
so basically it won't work for inputs like
2021
or 2021.1
and will error
so basically the "-" is optional in first test (since you can provide full version
2021.1.11-0.20220516.8e6c9917c-1 that will match in first test, but won't match in next one) , but a must match in second test, if first test won't work

so I don't know if above logic can be folded, if I'd fold it and use what you wrote it would continue even for 2021.1 and just pick more matches ... 
ok I guess the count of version test would then error out

ansible-scylla-node/tasks/Debian.yml
tarzanek: restore it back

tarzanek:
tested:
scylla_version: 2021.1.10
scylla_version: 2021.1.11
scylla_version: 2021.1.6-0.20211006.5eee352bf-1
scylla_version: latest

and negative (fails, since version is bad)
scylla_version: 2021.1
scylla_version: foobar

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -83,41 +83,24 @@
...
-
-        - name: Set scylla version if only one version with - found
-          set_fact:
-            sversion: "{{ rtempsversion }}"
-          when: rtempsversion | length == 1
-
-        - name: Error out, wrong Scylla version was passed, please fix it!
-          ansible.builtin.command: /bin/false
-          when: rtempsversion | length > 1
+    - name: Error out, wrong Scylla version was passed, please fix it!
```
vladzcloudius, line 100: You can specify why it is wrong - in this case because there is more than a single choice. Also, you don't seem to handle a case when `tempsversion | length == 0`.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/Debian.yml
vladzcloudius: I believe the easiest check is "rtempsversion | length != 1"

tarzanek: will fix

tarzanek:
amended, retested, @vladzcloudius any other comments?

ansible-scylla-node/tasks/Debian.yml
vladzcloudius: You forgot to address my "what's wrong" remark.
You can say something like "Error out, wrong Scylla version was passed (more than a single or none version matches), please fix it!"

vladzcloudius (changes requested):

vladzcloudius (changes requested):
Only a small nitpick about the error message. Except for that - looks good.

ansible-scylla-node/tasks/Debian.yml
tarzanek: fixed the error msg too now

vladzcloudius:
This is PR breaks the role.

Tests:
  * Install `2021.1.1` version (would fail because there would be multiple choices: 2021.1.1, 2021.1.11 and 2021.1.12).
  * If installed scylla version is not the latest and no upgrade is required: in this case `apt list -a` is going to print two entries that match the version pattern.

PR #135, opened 2022-06-30
Remove output to logs

jbragdon1:
This task seems to dump username / password / auth token to the outputted logs. Adding a no_log: true will still show the task running, but remove a great deal of the output.

Tested in a local copy

### DIFF
--- a/ansible-scylla-manager/tasks/add-clusters.yml
+++ b/ansible-scylla-manager/tasks/add-clusters.yml
@@ -22,3 +22,4 @@
       --password {{ item.password }} \
       {% endif %}
   when: cluster_already_added is not defined or cluster_already_added.stdout != "present"
+  no_log: true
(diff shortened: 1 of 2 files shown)


vladzcloudius:
Was going over the "deserted PRs" (sorry, @jbragdon1).
@vreniers this is an interesting one.

ansible-scylla-manager/tasks/add-clusters.yml
```diff
@@ -22,3 +22,4 @@
       --password {{ item.password }} \
       {% endif %}
   when: cluster_already_added is not defined or cluster_already_added.stdout != "present"
+  no_log: true
```
vladzcloudius, line 25: I'd parametrize it. And set it to `False` by default to preserve the current behavior.

vladzcloudius (changes requested):

ansible-scylla-manager/tasks/add-clusters.yml
vreniers, line 1: This PR requires a rebase. The file name was changed to `add_cluster.yml`.

vreniers:
I think we can make a broader issue out of this. 
The same happens with starting monitoring, or where cqlsh is used. We can hide those from the logs of course.

 I've opened an internal JIRA https://scylladb.atlassian.net/browse/ANSROLES-15 to track it. 

Since there was no reply or activity on this PR - I'll be closing it.

Issue #139, opened 2022-07-04
TLS - Private key (insecure) copy fails ansible is ran as a regular user


fee-mendes:
Seen at:

```shell
TASK [ansible-scylla-node : Copy the certificates into their proper locations] ************************************************************************************************************************************
changed: [172.31.23.8] => (item=./ssl/ca/testcluster-ca.crt)
changed: [172.31.23.8] => (item=./ssl/172.31.23.8/172.31.23.8.crt)
fatal: [172.31.23.8]: FAILED! => {"msg": "an error occurred while trying to read the file '/home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem': [Errno 13] Permission denied: b'/home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem'. [Errno 13] Permission denied: b'/home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem'"}
```

As private keys are created with `become: True` these will naturally default to the `root` user:

```shell
ubuntu@ip-172-31-15-172:~$ ls -l /home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem
-rw------- 1 root root 3243 Jul  4 01:10 /home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem
ubuntu@ip-172-31-15-172:~$ sudo file /home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem
/home/ubuntu/ssl/172.31.23.8/172.31.23.8.pem: PEM RSA private key
ubuntu@ip-172-31-15-172:~$ 
```

Therefore task `Copy the certificates into their proper locations` throws.

This issue should:

- No longer generate keys as root under localhost
- Ensure that private keys are copied with their correct permissions (600) as the convention is for many programs to fail when reading potentially insecure and exposed keys (Apache HTTPD & OpenSSH are some notable examples)

vreniers:
I'm closing this issue, it's stale. We're opening an internal JIRA to follow-up https://scylladb.atlassian.net/browse/ANSROLES-16.

PR #140, opened 2022-07-04
Node Role - Forbid Scylla upgrade when not explicitly requested

### LINKED ISSUE #132: [Scylla Node Role]: Refuse to upgrade unless "upgrade_version" is set to `true`
HEAD: 365a23dabb8fcd07a44654660847d37824e33a51

Node Role should refuse upgrading Scylla unless explicitly requested.
This behavior is meant preventing accidental upgrades.

Today if `upgrade_version: false` (default) Role will upgrade/downgrade Scylla bypassing the "upgrade" logic.

fee-mendes:
This series introduces an early check to determine whether Scylla is already installed on the target system.
It provides a mechanism to prevent incorrectly (and dangerously) upgrading a system which may be already part of a running cluster.

In such scenarios, we now require `upgrade_version: True`, which delegates ALL upgrade logic to its relevant upgrade section in the role.

To prevent a "chicken and egg" situation when we fail in the middle of a role execution AND after Scylla has been installed, we introduce the `skip_upgrade_check` option as a failsafe mechanism, which - obviously - defaults to false.

Tests ran:
1. Fail role execution after an install of Scylla in a single node cluster (this is - in fact, what led to https://github.com/scylladb/scylla-ansible-roles/issues/139 ;-)
2. Try to execute role again, we see a failure complaining that Scylla is already installed.
3. Add the `skip_upgrade_check: True` option, which allows us to get through and proceed with the installation/configuration process
4. **Revert** the `skip_upgrade_check: True` option, we see a failure complaining that Scylla is already installed.
5. Bump Scylla minor version and add the `upgrade_version: True` && `upgrade_major: False` options, upgrade completes.

Fixes #132

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -5,6 +5,12 @@
 # UPGRADE
 # ==============================
 
+# Skips role upgrade check. Useful if the role failed in the middle of an installation.
+# Values:
+# [default] false: Fail when Scylla is installed and the upgrade_version == False
+# true: When upgrade_version == False, re-run the install process 
+skip_upgrade_check: false
+
 # Upgrades Scylla version if a previous installation is detected
 # Values:
 #  [default] false: don't check for upgrades
--- a/ansible-scylla-node/tasks/main.yml
+++ b/ansible-scylla-node/tasks/main.yml
@@ -13,8 +13,9 @@
     msg: >
       Detected an existing Scylla installation.
       Should you want to upgrade, set `upgrade_version' to True.
+      If you are recovering from a failed role run, set `skip_upgrade_check' to True.
   when: >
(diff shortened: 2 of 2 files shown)


fee-mendes:
For some reason I can't add a reviewer -- so @vladzcloudius :-)

vladzcloudius (changes requested):
We can't merge it because we use a Role's state is not only scylla packages but rather a lot more: swap configuration, systemd configuration, additional packages and lot-lot more.

And sometimes we run a Role on a node/nodes with already installed Scylla in order to apply some changes to that state, e.g. to change `io_properties.yml` values.

Therefore the new logic this PR suggests is no-go.

What DO need to prevent unintentional upgrades and this IS the case today AFAIK.

PR #141, opened 2022-07-04
Ansible Node - Enforce private key permission and bind permissions to scylla user

### LINKED ISSUE #139: TLS - Private key (insecure) copy fails ansible is ran as a regular user

fee-mendes:
When the source playbook is executed with `become: True` its relevant TLS certificates will be owned by root. This causes a problem, because later on when trying to copy we are unable to read the resulting private key file, as it is (correctly) created by default with mode 0600.

This commit let Ansible generate each PKI component with mode 0644. Then, ensure these are copied to the remote machine's with strict permissions to scylla user/group. Finally, ensure that the private key is only readable by its target user - both locally and remotely.

Fixes #139

### DIFF
--- a/ansible-scylla-node/tasks/ssl.yml
+++ b/ansible-scylla-node/tasks/ssl.yml
@@ -38,6 +38,7 @@
     - name: Generate an OpenSSL private key.
       openssl_privatekey:
         path: "./ssl/{{ inventory_hostname }}/{{ inventory_hostname }}.pem"
+        mode: 0644
       delegate_to: localhost
 
     - name: Generate an OpenSSL Certificate Signing Request
@@ -45,6 +46,7 @@
         path: "./ssl/{{ inventory_hostname }}/{{ inventory_hostname }}.csr"
         privatekey_path: "./ssl/{{ inventory_hostname }}/{{ inventory_hostname }}.pem"
         common_name: "{{ inventory_hostname }}.{{ scylla_cluster_name }}.internal"
+        mode: 0644
       delegate_to: localhost
 
     - name: Generate an OpenSSL certificate signed with our CA certificate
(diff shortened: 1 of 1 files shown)


ansible-scylla-node/tasks/ssl.yml
```diff
@@ -54,21 +56,38 @@
...
+        src: "{{ item }}"
+        dest: "{{ scylla_ssl.cert_path }}/{{ item | basename }}"
+        owner: scylla
+        group: scylla
+        mode: '0600'
+      become: true
+      loop:
         - "./ssl/{{ inventory_hostname }}/{{ inventory_hostname }}.pem"
 
+    - name: Secure local OpenSSL private key
```
vladzcloudius, line 85: Why is this needed? Don't we want to cleanup after ourselves?

vladzcloudius (changes requested):

ansible-scylla-node/tasks/ssl.yml
fee-mendes: Makes sense. Updated to delete private key. Do notice we must still keep the private CA key for add-node operations, for instance.

vladzcloudius: I'm not following why would you need to keep any key after you copied it to remote nodes?
We are going to re-generate all keys from scratch every time we execute the Role anyway, don't we?

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -54,21 +56,38 @@
...
+        owner: scylla
+        group: scylla
+        mode: '0600'
+      become: true
+      loop:
         - "./ssl/{{ inventory_hostname }}/{{ inventory_hostname }}.pem"
 
+    - name: Delete local OpenSSL private key
+      file:
+        path: "./ssl/{{ inventory_hostname }}/{{ inventory_hostname }}.pem"
```
vladzcloudius, line 87: What's the point of keeping any file under .ssl on the localhost?

vladzcloudius (changes requested):

vreniers:
The reported problem still applies.

I've made an internal JIRA to follow-up on it: https://scylladb.atlassian.net/browse/ANSROLES-16

Since this PR has become stale, and the code to which it applies has changed somewhat. I'm closing this PR.

PR #143, opened 2022-07-04
OpenJDK - Fix inconsistencies

### LINKED ISSUE #142: Scylla install on Debian 10 fails as OpenJDK 8 is no longer supported
As seen in Debian 10:

```shell
TASK [ansible-scylla-node : Install Java] *************************************************************************************************************************************************************************
fatal: [181.10.10.125]: FAILED! => {"changed": false, "msg": "No package matching 'openjdk-8-jre-headless' is available"}
```

Details under https://lists.debian.org/debian-java/2019/04/msg00036.html

Resolution is to bump to JDK11. Problem: Exposes https://github.com/scylladb/scylla/issues/10442.

fee-mendes:
This series addresses two OpenJDK related inconsistencies across Ubuntu/Debian/CentOS:

As Debian 10 onward no longer ship with OpenJDK 8, we bump its version to OpenJDK 11. Similarly, we also let Ubuntu make use of that as it no longer ships updates to OpenJDK 8 (therefore, it should be eventually be removed in future releases). CentOS - on the other hand - still ships updates to OpenJDK 8 (via EPEL) and we let the default installer pick its latest release as usual.

However, installing a recent OpenJDK version in either distribution unveils https://github.com/scylladb/scylla/issues/10442, which needs to be treated accordingly. This is handled in the second commit in this series.

Fixes #142

### DIFF
--- a/ansible-scylla-node/tasks/Debian.yml
+++ b/ansible-scylla-node/tasks/Debian.yml
@@ -49,14 +49,14 @@
 
   - name: Install Java
     apt:
-      name: openjdk-8-jre-headless
+      name: openjdk-11-jre-headless
       state: present
       force_apt_get: yes
 
   - name: Correct java version selected
     alternatives:
       name: java
-      path: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
+      path: /usr/lib/jvm/java-11-openjdk-amd64/bin/java
 
   - name: "Going to install {{ scylla_version }}"
     set_fact:
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -258,9 +258,24 @@
         host: "{{ scylla_api_address }}"
         timeout: 300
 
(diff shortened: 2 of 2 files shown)


tarzanek:
@fee-mendes I think this patch looks good re update to openjdk 11

however the commit/patch for nodetool and tainting environment with java opts should be split from this PR and likely I don't want to merge it (it's fixed in latest Scylla anyways, no need to pollute the repo with temporary stuff that is not even relevant now)

vladzcloudius:
In addition to what @tarzanek wrote - is it even safe to use OpenJDK11? Does nodetool work with anything but Java 1.8?

@amnonh comments?

fee-mendes:
> is it even safe to use OpenJDK11? Does nodetool work with anything but Java 1.8?

If it isn't then we have a real problem in our release process. 1.8 is EOL for a long time now, most distributions support it on a best-effort basis, and Debian - as always progressist - already removed it. :-)

PR #145, opened 2022-07-06
Node Role - upgrade_major should default to False

fee-mendes:
It makes no sense for `upgrade_major` to be set to `True`. Major upgrades must **always** be explicitly requested, whereas minor ones should always have the precedence.


vladzcloudius (approved):
Makes sense

PR #153, opened 2022-08-18
scylla 2022 upgrade, monitoring 4.0, manager 3.0

tarzanek:


vladzcloudius (changes requested):
Please, put a little more detailed description for the patch.
In particular, describe the motivation for it.

tarzanek:
will fix the message
anyhow the bigger problem is the failing test

because molecule stopped working for newer Scylla versions, so we diverged somewhere and there are no logs,
seems we need a way to make molecule show more logs about service failure
```
 TASK [ansible-scylla-node : start scylla on node2] *****************************
  fatal: [node2]: FAILED! => {"changed": false, "msg": "Unable to start service scylla-server: Job for scylla-server.service failed because a fatal signal was delivered causing the control process to dump core.\nSee \"systemctl status scylla-server.service\" and \"journalctl -xe\" for details.\n"}
  ```

PR #159, opened 2022-09-06
Simplify rolling restart for 2021 and up Scylla versions with seedless

tarzanek:


vladzcloudius:
Why don't you fix the existing playbook?

vladzcloudius (changes requested):
Why don't you fix the existing playbook?

PR #162, opened 2022-09-09
Node role - Correctly generate CA certificate and update references

### LINKED ISSUE #161: Ansible Role - No v3_ca extension in CA certificates
As it turns out, for [this SO](https://stackoverflow.com/questions/53881437/invalid-ca-certificate-with-self-signed-certificate-chain/53951346#53951346) posting OpenSSL [let you sign other certs with a non-CA root cert](https://superuser.com/a/464288/778853)  but verification of such certs will fail (because the CA check will fail).

Certificate Authorities need to come with the `v3_ca` extension, which - when inspecting the inner certificate content - shall report that the certificate in question **IS** a CA:

```shell
$ openssl x509 -in cert.crt -text| grep 'CA:'
                CA:TRUE, pathlen:0
```

However, the Ansible Role today will fail to create such a CA:
```shell
$ openssl x509 -in ansible-ca.crt -text| grep 'CA:' 
$ 
```

This happens because we don't tell OpenSSL which kind of certificate we are generating - which in turn - makes it invalid for later use:
https://github.com/scylladb/scylla-ansible-roles/blob/dd3ef330a13d773b705ca3d531fbb80e9312862f/ansible-scylla-node/tasks/ssl.yml#L13:L17

Note that the certificate in question has no special meaning and no special extensions, which means that we are actually telling OpenSSL to generate a regular certificate.

Moreover, another place where such situation is confirmed is when we analyze the `cqlshrc` template, which instead of referencing the CA file, will reference the **node** generated certificate with a `validate = False`, which means the CA is not really used for anything at this point.

https://github.com/scylladb/scylla-ansible-roles/blob/dd3ef330a13d773b705ca3d531fbb80e9312862f/ansible-scylla-node/templates/cqlshrc.j2#L10:L11

The impact of this is that users won't be able to connect via a TLS connection unless they decide to skip the certificate verification check. If they offer the CA as a certificate and verify it, the result will be:

>> cassandra.cluster.NoHostAvailable: ('Unable to connect to any servers', {'X.X.X.X:9142': PermissionError(1, "Tried connecting to [('X.X.X.X', 9142)]. Last error: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: invalid CA certificate (_ssl.c:997)")})

@vladzcloudius @tarzanek

fee-mendes:
Our previous implementation of ssl.yml incorrectly generated a self-signed certificate mistakenly named as a "CA".

As it turns out, a CA certificate needs to have some specific characteristics in order to be considered a proper Certificate Authority.
Due to that, the CA certificate couldn't be used by clients to properly authenticate against the nodes, in such a way that clients were required to skip TLS verification in order to connect.

This mini-series corrects that situation and also updates the generated `cqlshrc` template to properly use the CA certificate created.

Fixes #161 
Closes #161

### DIFF
--- a/ansible-scylla-node/tasks/ssl.yml
+++ b/ansible-scylla-node/tasks/ssl.yml
@@ -11,18 +11,23 @@
         path: "./ssl/ca/{{ scylla_cluster_name }}-ca.pem"
 
     - name: Generate an OpenSSL Certificate Signing Request for the CA
-      openssl_csr:
-        path: "./ssl/ca/{{ scylla_cluster_name }}-ca.csr"
+      community.crypto.openssl_csr_pipe:
         privatekey_path: "./ssl/ca/{{ scylla_cluster_name }}-ca.pem"
         common_name: "{{ scylla_cluster_name }}.internal"
+        use_common_name_for_san: false  # since we do not specify SANs, don't use CN as a SAN
+        basic_constraints:
+          - 'CA:TRUE'
+        basic_constraints_critical: yes
+        key_usage:
+          - keyCertSign
+        key_usage_critical: true
+      register: ca_csr
 
     - name: Generate a Self Signed OpenSSL certificate for the CA
-      openssl_certificate:
+      community.crypto.x509_certificate:
         path: "./ssl/ca/{{scylla_cluster_name }}-ca.crt"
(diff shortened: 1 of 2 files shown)


tarzanek (approved):
lgtm

PR #165, opened 2022-09-13
enable service for node exporter for 5.0 and 2022

tarzanek:


tarzanek:
merging, despite the missing check, it's better to have this working with 2022
we can fix the proper check for Scylla version later

PR #167, opened 2022-09-13
Adds "rolling decomission" playbook

ebenzecri:

### DIFF
--- a/ansible-scylla-monitoring/tasks/Debian.yml
+++ b/ansible-scylla-monitoring/tasks/Debian.yml
@@ -40,7 +40,7 @@
 
   - name: add repo
     shell: |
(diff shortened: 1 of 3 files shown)


ebenzecri:
Specific tests done against this playbook:

| Scenario | Results |
|---|---|
| All nodes are in UN status | All selected nodes were successfully decommissioned |
| A node from the cluster in UL status | Playbook will wait until the node finally leaves the cluster to start with the next one from the list |
| A node from the decommissioning list is in DN status | Playbook will fail because it can't use the API |
| A node from the decommissioning list has `scylla-server` service is stopped | Playbook will fail because it can't use the API |
| A node from the decommissioning list was already decommissioned | Playbook will ignore the node (only if `scylla-server` is still running) |

ebenzecri:
Please, close this PR. A new PR related to https://github.com/scylladb/scylla-ansible-roles/issues/171 will be opened. Thanks!

PR #170, opened 2022-09-22
Ansible Node: Use token_distributor.py to generate tokens

igorribeiroduarte:


ansible-scylla-node/tasks/start_one_node.yml
```diff
@@ -1,10 +1,109 @@
...
+
+- name: Replication aware token distributor tasks
+  block:
+    - name: Get current token ring
+      block:
+        - name: Get tokens
+          shell: |
+            for ip in `nodetool status | grep -Eo '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}'`; do
+                echo "$ip=`curl -s -X GET http://localhost:10000/storage_service/tokens/$ip | cut -d '[' -f2 | cut -d ']' -f1 | tr -d \\" | tr -d \ `"
+            done
```
vladzcloudius, line 18: This looks terribly over complicated and messy.
Why couldn't you:
1) Calculate tokens for all nodes in the cluster (based on the inventory.ini) and define a per-host list that contains these tokens.
2) Leave the Role's code that bootstraps nodes as it does today except for:
  * If node is actually being bootstrapped (**) you:
    * Set it pre-calculated tokens using `initial_token` parameter
    * Bootstrap.
    * Remove `initial_token`.

(**) to identify which nodes need to be bootstrapped you can use a `nodetool ring` or the corresponding REST API (preferably) to fetch all known nodes and their current state:
 * If there is any node not in a "NORMAL" state - exit with an error.

This needs to be executed on any alive node: alive nodes can be checked using CQL port probing.

And all this needs to be done before the bootstrapping loop and NOT in `start_one_node.yml`.

`start_one_node.yml` needs to use already pre-calculated variables and simply act on them as described above.

vladzcloudius (changes requested):
This looks very wrong.
Let's have a design document first describing what and how you want to do things.

ansible-scylla-node/defaults/main.yml
```diff
@@ -260,6 +260,13 @@ skip_sysconfig: False
 skip_selinux: False
 skip_ntp: False
 
+# If defined, this enables the usage of a separate token_distritor instead of scylla's default random token distributor
```
sharonovd, line 263: Is @tarzanek ok with such a way of handling dependencies? I am not proficient enough in ansible to suggest a better one, unfortunately

sharonovd:
This functionality should be decribed in the doc (or wiki). Basically, you're introducing new pluggable functionality, so please think better on defining it's plug-in API. F
or example, I'd expect that I have to provide a path to my own distributor script, but then:
 - field_cloud_access path makes no sense, this is something to be handled by token_distributor.py should handle it itself
 - bried description of distributor's API is necessary.

igorribeiroduarte:
> This functionality should be decribed in the doc (or wiki). Basically, you're introducing new pluggable functionality, so please think better on defining it's plug-in API. For example, I'd expect that I have to provide a path to my own distributor script, but then:
> 
>     * field_cloud_access path makes no sense, this is something to be handled by token_distributor.py should handle it itself)
> 
>     * bried description of distributor's API is necessary.

Done, @sharonovd

ansible-scylla-node/defaults/main.yml
tarzanek: so I don't think there are good ways to achieve this, let me try to find a way, but for now documenting is better than nothing

tarzanek: typo in 
" token_distritor "

igorribeiroduarte: Fixed

tarzanek:
test result with custom distributor:
```
XXX@test-token-distributor2-node-us-central1-a-0:~$ nodetool status quotes
Datacenter: us-central1
=======================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address      Load       Tokens       Owns (effective)  Host ID                               Rack
UN  10.128.0.33  1.21 MB    256          49.9%             8ae45715-7585-430e-99de-a3327d45037a  a
UN  10.128.0.34  1.05 MB    256          50.1%             eb889417-d1ea-48eb-b1ac-afb6237f13e8  a
UN  10.128.0.35  1.49 MB    256          50.0%             9a82ce84-4c2a-42da-84e4-016c564859ef  b
UN  10.128.0.36  1.57 MB    256          50.0%             b3cbeeda-a40b-41c8-99e9-68a7fee3e39c  b
UN  10.128.0.37  1.31 MB    256          50.0%             0ab273dd-ff61-411a-8279-9e0ebf3a4dbc  c
UN  10.128.0.38  1.64 MB    256          50.0%             26f46916-0d28-49d7-9bcf-fac246375984  c
```

Issue #169, opened 2022-09-19
Cleanup playbook runs in serial mode


ebenzecri:
Since `serial: 1` is hard-coded, that option must be able to be overridden, setting its value to a variable.

PR #172, opened 2022-10-03
example-playbooks: Implements batch size

ebenzecri:
This change allows the playbook to be run on multiple nodes at the same time

Ref: #169


tarzanek:
ok, let's merge it, people need to however seriously understand what they are doing, since this will impact performance of cluster if it will run on more than 1 node at the same time

tarzanek:
just a ref: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_strategies.html#setting-the-batch-size-with-serial

PR #175, opened 2022-11-03
Role ansible-scylla-manager: correct default variables

v0112358:
correct default variables


tarzanek:
thnx for fixing this typo @v0112358 , merging

PR #182, opened 2022-11-30
[#180] Switch over to lt kernel and remove old kernels

### LINKED ISSUE #180: Issues with kernels on CentOS hosts
I have been speaking with @tarzanek on slack as I had hit some issues with the kernel-ml package that (when I started using this repo) was installed by default

I observed:
 - with kernel-ml > 5.x (from around August this year) nvme drives in AWS were not detected on my instances
 - on hosts that had kernel-ml installed in the past that had later been updated using ansible, had new default kernels (3.10) installed which became grub option 0 and were booted in to

To fix this, I did:
 - used kernel-lt over kernel-ml
 - permanently enabled the epel-kernels repo
 - removed the kernel package

Rumbles:
Extends #181 to enable the kernel repo, remove old kernels and set the machine to reboot if required by yum

Closes #180

### DIFF
--- a/ansible-scylla-node/tasks/RedHat.yml
+++ b/ansible-scylla-node/tasks/RedHat.yml
@@ -26,13 +26,28 @@
         name: "https://www.elrepo.org/elrepo-release-{{ ansible_facts['distribution_major_version'] }}.el{{ ansible_facts['distribution_major_version'] }}.elrepo.noarch.rpm"
         state: present
 
+    - name: enable epel kernels
+      shell: |
+        yum-config-manager --enable elrepo-kernel
+
     - name: Install ELRepo mainline kernel
       yum:
         name: kernel-lt
         state: present
-        enablerepo: elrepo-kernel
         lock_timeout: 60
       register: kernel_installed
+
+    - name:
+      package:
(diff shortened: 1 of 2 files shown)


ansible-scylla-node/tasks/RedHat.yml
```diff
@@ -62,7 +77,7 @@
         pre_reboot_delay: 0
         post_reboot_delay: 30
         test_command: whoami
-      when: reboot_after_kernel_upgrade|bool and kernel_installed.changed
+      when: (reboot_after_kernel_upgrade|bool and kernel_installed.changed) r restart_required.rc == 1
```
tarzanek, line 80: OR ? (not r ? ) typo @Rumbles ?

Rumbles: Hi, you're right there is a mistake here, but I ran in a to a larger issue... when I tested this method, the yum package absent step will fail as one of the kernels it tries to remove is the running kernel...

Rumbles: ok, I think that should be better now

Issue #9, opened 2021-01-13
Authenthication enabling should adjust replication of system_auth keyspace


tarzanek:
default system_auth is set to RF=2  with simple strategy
for a cluster this should be set to NetworkTopology + RF = number of nodes and trigger a repair
otherwise every auth connect triggers connections around the cluster and not local ones

dyasny:
This is true, but schema changes require CQL access, and that would mean the playbooks will have to be fed credentials. I think it would make much more sense to move all CQL related automation to a separate project (a pile of scripts really) that uses a Scylla driver and does the right things there. Those can be triggered by the roles in turn of course.

mykaul:
@vladzcloudius - thoughts?

vladzcloudius:
We can implement it for now.

PR #193, opened 2023-01-16
ansible-scylla-node: Verify if auth is enabled and adjust system_auth ks

igorribeiroduarte:
Closes https://github.com/scylladb/scylla-ansible-roles/issues/9

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -126,6 +126,8 @@ full_inventory: true
 # !!! also make sure you override scylla_yaml_params and remove authorization from there !!!
 scylla_authentication: True
 
+repair_timeout_seconds: 36000
+
 # set to enabled when the scylla-server service should be started by applying the role.
 # if, for example, we add a node, but prefer to wait before we start the service and add it to the cluster,
 # this should be set to false
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -126,7 +126,19 @@ full_inventory: true
 # !!! also make sure you override scylla_yaml_params and remove authorization from there !!!
 scylla_authentication: True
 
+# Set this to false to disable automatic adjustment of system_auth keyspace replication
+adjust_system_auth_replication: False
+
+# Replication variables and cql credentials are intended to be used to adjust system_auh keyspace when necessary.
+# Remember to update cql credentials here in case you're making updates to an already existing cluster which
+# doesn't have the default username/password(cassandra/cassandra)
+system_auth_rf: 3
+system_auth_replication_strategy: NetworkTopologyStrategy
+cql_username: cassandra
+cql_password: cassandra
+
 repair_timeout_seconds: 36000
+cleanup_timeout_seconds: 36000
 
 # set to enabled when the scylla-server service should be started by applying the role.
 # if, for example, we add a node, but prefer to wait before we start the service and add it to the cluster,
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -252,21 +252,21 @@
     - (scylla_ssl.internode.enabled|bool) or
       (scylla_ssl.client.enabled|bool)
 
+# scylla_listen_address is a composite indirect value that depends on another per-host value - scylla_nic.
+# Therefore in order to be able to get the corresponding value via hostvars[item] later in the play we need to
+# have an actual value resolved.
+# One way to do that is to resolve it and store as an Ansible fact.
+- name: Resolve a scylla_listen_address as a fact
+  set_fact:
+    listen_address: "{{ scylla_listen_address }}"
+
+# The same relates to the below
+- name: Resolve scylla_broadcast_address
+  set_fact:
+    broadcast_address: "{{ scylla_broadcast_address }}"
+
 - name: start_scylla_service dependent tasks
   block:
-    # scylla_listen_address is a composite indirect value that depends on another per-host value - scylla_nic.
-    # Therefore in order to be able to get the corresponding value via hostvars[item] later in the play we need to
-    # have an actual value resolved.
-    # One way to do that is to resolve it and store as an Ansible fact.
-    - name: Resolve a scylla_listen_address as a fact
-      set_fact:
-        listen_address: "{{ scylla_listen_address }}"
-
-    # The same relates to the below
-    - name: Resolve scylla_broadcast_address
-      set_fact:
-        broadcast_address: "{{ scylla_broadcast_address }}"
-
     - name: Generate tokens for the new nodes
       run_once: true
       include_tasks: generate_tokens.yml
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -329,6 +329,75 @@
       when: token_distributor is defined
   when: start_scylla_service is defined and start_scylla_service|bool
 
+- name: Check if authentication is enabled
+  block:
+    - command: cat /etc/scylla/scylla.yaml
+      ignore_errors: true
+      register: _scylla_yaml_out
+
+    - set_fact:
+        _scylla_yaml_map: "{{ _scylla_yaml_out.stdout | from_yaml }}"
+
+    - set_fact:
+        _authentication_enabled: |
+          {% if (_scylla_yaml_map.authenticator is defined and (_scylla_yaml_map.authenticator == 'TransitionalAuthenticator' or _scylla_yaml_map.authenticator == 'PasswordAuthenticator')) or
+          (_scylla_yaml_map.authorizer is defined and (_scylla_yaml_map.authorizer == 'TransitionalAuthorizer' or _scylla_yaml_map.authorizer == 'CassandraAuthorizer')) %}True{% else %}False{% endif %}
+  run_once: true
+
+- name: If authentication is enabled, adjust replication for system_auth keyspace
+  block:
+    - name: Validate that all nodes are up before adjusting the replication
+      uri:
+        url: "http://{{ scylla_api_address }}:{{ scylla_api_port }}/failure_detector/endpoints/"
+        follow_redirects: none
+        method: GET
+      register: _result
+      until: _result.status == 200
+      retries: 10
+      delay: 1
+
+    - name: Get datacenter name
+      uri:
+        url: "http://{{ scylla_api_address }}:{{ scylla_api_port }}/snitch/datacenter"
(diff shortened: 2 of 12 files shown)


ansible-scylla-node/tasks/repair.yml
```diff
@@ -0,0 +1,23 @@
...
+  retries: 5
+  delay: 1
+
+- name: Wait for the repair to finish
+  uri:
+    url: "http://{{ scylla_api_address }}:{{ scylla_api_port }}/storage_service/repair_async/{{ keyspace }}?id={{ _repair_id.json }}"
+    method: GET
+  register: _repair_status
+  until: _repair_status.json != 'RUNNING'
+  retries: 1200 # 1200 * 30 seconds = 10 hours
```
vladzcloudius, line 17: Let's parametrize it: create a variable that would hold a corresponding delay and create a reasonable default (e.g. 10h as here).

ansible-scylla-node/tasks/repair.yml
```diff
@@ -0,0 +1,23 @@
+---
+- name: Start repair
```
vladzcloudius, line 2: Please, add an inline description explaining that this does exactly: it repairs primary token range belonging the node with a REST served from `scylla_api_address`.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/common.yml
```diff
@@ -299,6 +299,45 @@
       when: token_distributor is defined
   when: start_scylla_service is defined and start_scylla_service|bool
 
+- name: Check if authentication is enabled
+  lineinfile:
+    path: /etc/scylla/scylla.yaml
+    regexp: '^(authenticator: (.*PasswordAuthenticator.*|.*TransitionalAuthenticator.*))'
```
vladzcloudius, line 305: This is not enough - `authorizer` value needs to be checked too.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/common.yml
```diff
@@ -299,6 +299,45 @@
...
+      run_once: true
+
+    - name: Repair system_auth
+      include_tasks: repair.yml
+      vars:
+        keyspace: 'system_auth'
+
+    - name: Cleanup system_auth
+      uri:
+        url: "http://{{ scylla_api_address }}:{{ scylla_api_port }}/storage_service/keyspace_cleanup/system_auth"
```
vladzcloudius, line 331: Isn't this API blocking?

vladzcloudius (changes requested):

ansible-scylla-node/tasks/repair.yml
igorribeiroduarte: Done

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Fixed

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Yes, but now I'm using the builtin ansible async task

igorribeiroduarte: Fixed

ansible-scylla-node/tasks/common.yml
```diff
@@ -299,6 +299,59 @@
...
+      shell: |
+        cqlsh {{ broadcast_address }} -u {{ cql_username }} -p {{ cql_password }} -e "ALTER KEYSPACE system_auth WITH replication = {'class': '{{ replication_strategy }}', {{ dcs_to_rf | unique | join(',') }}};"
+      run_once: true
+
+    - name: Repair system_auth
+      include_tasks: repair.yml
+      vars:
+        keyspace: 'system_auth'
+
+    - name: Cleanup system_auth
```
vladzcloudius, line 338: In general it's more efficient to cleanup first and then repair.
This way a repair will have less data in sstables to go through.

vladzcloudius (changes requested):
Looks good in general except for a small nitpick.

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Fixed

ansible-scylla-node/defaults/main.yml
```diff
@@ -120,7 +120,19 @@ full_inventory: true
 # !!! also make sure you override scylla_yaml_params and remove authorization from there !!!
 scylla_authentication: True
 
+# Set this to false to disable automatic adjustment of system_auth keyspace replication
```
vladzcloudius, line 123: Please, add the description of every new variable this patch adds to a patch description.

Also, please, clarify the motivation for this patch in the patch description too (start with it).

ansible-scylla-node/defaults/main.yml
```diff
@@ -120,7 +120,19 @@ full_inventory: true
...
+adjust_system_auth_replication: False
+
+# Replication variables and cql credentials are intended to be used to adjust system keyspaces when necessary
```
vladzcloudius, line 126: what "system keyspaces" exactly? A `system` KS is a node local for example.

ansible-scylla-node/defaults/main.yml
```diff
@@ -120,7 +120,19 @@ full_inventory: true
...
+# Remember to update cql credentials here in case you're making updates to an already existing cluster which
+# don't have the default username/password(cassandra/cassandra)
+replication_factor: 3
```
vladzcloudius, line 129: This variable name is not good - too generic. What if we need to define a replication factor for some other KS(es) in the future?

Something like `scylla_internal_kses_replication_factor` would be better IMO.

ansible-scylla-node/defaults/main.yml
```diff
@@ -120,7 +120,19 @@ full_inventory: true
...
+replication_strategy: NetworkTopologyStrategy
```
vladzcloudius, line 130: The same as with `replication_factor` - too generic.

ansible-scylla-node/tasks/common.yml
```diff
@@ -299,6 +299,63 @@
       when: token_distributor is defined
   when: start_scylla_service is defined and start_scylla_service|bool
 
+- name: Check if authentication is enabled
+  lineinfile:
+    path: /etc/scylla/scylla.yaml
+    regexp: '^(authenticator: (.*PasswordAuthenticator.*|.*TransitionalAuthenticator.*))|(authorizer: (.*CassandraAuthorizer.*|.*TransitionalAuthorizer.*))'
```
vladzcloudius, line 305: Huh?
Why do you have to process a YAML file as a plain text one?
Doesn't Ansible have YAML parsers for that?

vladzcloudius: And it turns out it does: https://stackoverflow.com/questions/68217313/parse-yaml-files-in-ansible

ansible-scylla-node/defaults/main.yml
vladzcloudius: But what if tomorrow we would want a different config for `audit` and `system_auth`?

So, better define different variables for each.

ansible-scylla-node/tasks/common.yml
```diff
@@ -299,6 +299,63 @@
...
+    state: absent
+  check_mode: yes
+  register: _authentication_enabled
+  run_once: true
+
+- name: If authentication is enabled, adjust replication for system_auth keyspace
+  block:
+    - fail:
+        msg: "Unable to adjust system_auth replication unless start_scylla_service is set to True"
+      when: start_scylla_service == False
```
vladzcloudius, line 315: What if you were not supposed to start the service but it was already running?
You should check if `scylla` is running instead.

ansible-scylla-node/tasks/common.yml
```diff
@@ -299,6 +299,63 @@
...
+        shell: |
+          nodetool cleanup system_auth
+        alias: scylla_cleanup_system_auth
+        async: "{{ cleanup_timeout_seconds }}"
+        retries: "{{ cleanup_timeout_seconds // 30 }}" # retries = cleanup_timeout_seconds / delay
+        delay: 30
+      register: _cleanup_output
+
+    - name: Cleanup logs
+      debug: var=_cleanup_output
```
vladzcloudius, line 348: What is this about?

ansible-scylla-node/templates/cqlshrc.j2
```diff
@@ -1,6 +1,11 @@
 [authentication] ;; Change these after the user/password has been updated from the default
+{% if cql_username is defined and cql_password is defined %}
+username = {{ cql_username }}
+password = {{ cql_password }}
+{% else %}
```
vladzcloudius, line 5: How this patch is related to this PR?

vladzcloudius (changes requested):

ansible-scylla-node/defaults/main.yml
igorribeiroduarte: Done

ansible-scylla-node/defaults/main.yml
igorribeiroduarte: Done

ansible-scylla-node/defaults/main.yml
igorribeiroduarte: Fixed

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Fixed

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Fixed

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Just watching the cleanup logs, in the same way we do for the `run_decomission` and `run_cleanup` playbooks.

ansible-scylla-node/templates/cqlshrc.j2
igorribeiroduarte: At first I though it would be a good idea to use these new credentials to update cqlshrc as well, but now I see that it doesn't make sense.
Patch removed.

vladzcloudius (changes requested):
This PR fails the CI test for a bad reason - there is a real failure.
Please, fix and it's good to be merged.

igorribeiroduarte:
> This PR fails the CI test for a bad reason - there is a real failure. Please, fix and it's good to be merged.

Done

PR #195, opened 2023-01-17
scylla_node: don't ignore errors when setting up a raid

### LINKED ISSUE #152: recognize various situations for raid_setup and stop in appropriate cases
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/tasks/common.yml#L35

wasn't idempotent, hence the errors are ignored

this block needs to be made idempotent at least for cases where raid exists, so it won't try to recreate it and continue
and of course it should fail for other reasons

vladzcloudius:
We should never ignore errors in general and when setting the RAID up in particular.

If RAID exists, and it doesn't correspond to the configuration required - this requires an Operator intervention.

Fixes #152

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -32,7 +32,6 @@
         scylla_raid_setup --disks "{{ scylla_raid_setup | join(',') }}" --update-fstab
       become: true
       when: scylla_raid_setup is defined and scylla_raid_setup|length > 0 and scylla_raid_setup|length != present_raid_disks.stdout|int
-      ignore_errors: true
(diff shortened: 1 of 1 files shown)


vladzcloudius:
Tested it manually.

tarzanek:
did you test also Update Cluster?
(resp. reruning the role on top of existing cluster?)

if it works properly, then let's merge this

tarzanek:
I guess we should get a way from operator to override this check instead of making it strict
- meaning by default it will be as you just changed it to
but operator could get a variable that can let him decide without role change to override and continue despite this check/create will fail

PR #198, opened 2023-01-20
extra-playbook: Ubuntu kernel pin for GCP

ebenzecri:


vladzcloudius:
@ebenzecri What's the motivation for this PR and how are you achieving it?
In particular it's not clear to me why this is not a part of the Scylla Node Role. It has a code that installs RH Repl kernels. How is this different?

mykaul:
Would be good to extract the stop scylla and start scylla tasks from this, as I assume this are useful tasks that we use elsewhere as well?

PR #200, opened 2023-01-24
molecule: upgrade ansible to 6.7.0

tarzanek:


tarzanek:
@sharonovd still works on this, so while this fixes dependencies, it lacks fix for containers

tarzanek:
@sharonovd can you review or merge if you're ok with this?

sharonovd:
Note to self - we need to switch CI to self-hosted runner, because public runners don't allow this style of dind for some reason now.

sharonovd (approved):

tarzanek:
let's move the containers not being able to start to a different issue then

Issue #116, opened 2022-04-20
[ansible-scylla-node]: make scylla_io_probe DC-aware


vladzcloudius:
*Description*
Homogeneous HW configuration requirement (assumption) makes sense in a scope of a single DC.
However different DCs can have different HW configuration starting with I/O backend.

Let's make scylla_io_probe DC-aware: when set to True:
* For each DC:
  1) Choose the first node from the inventory in that DC.
  2) Configure I/O on that node.
  3) Set I/O configuration on each node in the current DC to be the same as on the node from (1).

vladzcloudius:
This is related to that new format in `inventory.ini` I mentioned.
@tarzanek if you are going to pick it up - please, assign yourself and unassign Igor.

PR #217, opened 2023-02-01
ansible-scylla-node: Make scylla_io_probe DC aware

igorribeiroduarte:
Closes https://github.com/scylladb/scylla-ansible-roles/issues/116

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -175,6 +175,12 @@ gpfs_prefer_local: "true"
 # probe for IO figures on one node and reuse the figures on all nodes (recommended: True)
 scylla_io_probe: True
 
+# If set to True, scylla_io_probe must be True and dc variable must be defined in the inventory
+# for each one of the nodes.
+# This will make sure that scylla_io_setup will be executed in one node from each datacenter
+# and then the configuration files will be propagated for the other nodes of each datacenter.
+scylla_io_probe_dc_aware: False
+
 # These can be arbitrarily set if scylla_io_probe is set to False
 # io_properties:
 #   disks:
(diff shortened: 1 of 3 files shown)


vladzcloudius (changes requested):
Please, add a proper patch description.

ansible-scylla-node/defaults/main.yml
```diff
@@ -175,6 +175,12 @@ gpfs_prefer_local: "true"
 # probe for IO figures on one node and reuse the figures on all nodes (recommended: True)
 scylla_io_probe: True
 
+# If set to True, scylla_io_probe must be True and dc variable must be defined in the inventory
+# for each one of the nodes.
+# This will make sure that scylla_io_setup will be executed in one node from each datacenter
+# and then the configuration files will be propagated for the other nodes of each datacenter.
+scylla_io_probe_dc_aware: False
```
vladzcloudius, line 182: Why false by default?

vladzcloudius (changes requested):
Looks good in general. 
One small comment about the default and a proper description and it is good to be merged.

Nice work!

ansible-scylla-node/defaults/main.yml
igorribeiroduarte: My concern is that for any already existing cluster which don't use `gossipingPropertyFileSnitch`, if this new variable is set to `True` by default, as soon as we run an update task in any of these clusters, this will update the `io_properties` for this cluster and in this case the `DC` variable might not be present, what would cause the playbook execution to fail, or the `DC` variable might have wrong information, and in this case the playbook wouldn't fail but we would end up with wrong io settings.

igorribeiroduarte: Patch description added

vladzcloudius (changes requested):
Typo in the description: s/scylla_io_proce_dc_aware/scylla_io_probe_dc_aware/

igorribeiroduarte:
> Typo in the description: s/scylla_io_proce_dc_aware/scylla_io_probe_dc_aware/

Fixed

vladzcloudius (approved):

PR #222, opened 2023-02-13
monitoring: add support for monitoring 4.x with multiple prometheus f…

### LINKED ISSUE #221: add support for monitoring 4.x with multiple prometheus files
newer prometheus in monitoring delivers:
```
ls scylla-monitoring-scylla-monitoring-4.2.1/prometheus/prom_rules/
back_fill  prometheus.latency.rules.yml  prometheus.rules.yml  prometheus.table.yml
```

a directory based approach to rules files

while role forces only single file:
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-monitoring/tasks/docker.yml#L39
( copied by https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-monitoring/tasks/common.yml#L71 )

we need to switch to using a directory as newer monitoring too
(this approach will be backwards compatible, so it's just about improving the role)

tarzanek:
…iles

fixes #221

### DIFF
--- a/ansible-scylla-monitoring/defaults/main.yml
+++ b/ansible-scylla-monitoring/defaults/main.yml
@@ -22,6 +22,7 @@ scylla_monitoring_archive_url: 'https://github.com/scylladb/scylla-monitoring/ar
 scylla_monitoring_deploy_path: /opt/scylla-monitoring
 scylla_monitoring_data_path: "{{ scylla_monitoring_deploy_path }}/data"
 scylla_monitoring_config_path:  "{{ scylla_monitoring_deploy_path }}/config"
+scylla_monitoring_prom_rules_path:  "{{ scylla_monitoring_config_path }}/prom_rules"
 scylla_monitoring_alertdata_path: "{{ scylla_monitoring_deploy_path }}/alertdata"
 
 # List of monitoring dashboards to enable
@@ -33,7 +34,8 @@ scylla_monitoring_dashboards_versions:
 # If a path is provided, the file(s) will be taken from a local path and pushed to the config directory for the monitoring installation.
 # If "default" is set, the default file will be taken from the prometheus directory in the unpacked monitoring stack distribution.
 #
-# See the original default file at https://github.com/scylladb/scylla-monitoring/blob/master/prometheus/prom_rules/prometheus.rules.yml
+# Sample default file was prometheus.rules.yml at https://github.com/scylladb/scylla-monitoring/blob/master/prometheus/prom_rules/
+# below file if other than default will be added to files from above prom_rules folder as "userdefined.rules.yml"
 prometheus_rules_file: default
 # See the original default file at https://github.com/scylladb/scylla-monitoring/blob/master/prometheus/prometheus.yml.template
 prometheus_yml_template_file: default
--- a/ansible-scylla-monitoring/tasks/common.yml
+++ b/ansible-scylla-monitoring/tasks/common.yml
@@ -17,9 +17,28 @@
     owner: "{{ ansible_user_id }}"
   become: true
 
+- name: change permissions for the data directory to nobody
+  file:
+    path: "{{ scylla_monitoring_data_path }}"
+    state: directory
+    mode: '1777'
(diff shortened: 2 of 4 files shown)


ansible-scylla-monitoring/defaults/main.yml
```diff
@@ -34,6 +35,7 @@ scylla_monitoring_dashboards_versions:
 # If "default" is set, the default file will be taken from the prometheus directory in the unpacked monitoring stack distribution.
 #
 # See the original default file at https://github.com/scylladb/scylla-monitoring/blob/master/prometheus/prom_rules/prometheus.rules.yml
+# above file will STILL be used, below file if other than default will be added to above file as addition (along to all other files inside prom_rules)
```
sharonovd, line 38: I struggle to understand the sentence, read it three times :(

sharonovd (commented):
One minor comment

ansible-scylla-monitoring/defaults/main.yml
tarzanek: rewritten and amended

ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -71,7 +90,7 @@
 - name: set prometheus rules file from preset file
   copy:
     src: "{{ prometheus_rules_file }}"
-    dest: "{{ scylla_monitoring_config_path }}/prometheus.rules.yml"
+    dest: "{{ scylla_monitoring_prom_rules_path }}/userdefined.rules.yml"
```
vladzcloudius, line 93: What about older Monitoring versions?
Wouldn't this effectively break their support?

ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -93,13 +112,13 @@
...
+- name: copy ALL rules files from the default prom_rules file location
   copy:
-    src: "{{ base_dir }}/prometheus/prom_rules/prometheus.rules.yml"
+    src: "{{ base_dir }}/prometheus/prom_rules/*"
     remote_src: true
-    dest: "{{ scylla_monitoring_config_path }}/prometheus.rules.yml"
+    dest: "{{ scylla_monitoring_prom_rules_path }}/"
     mode: '0644'
-  when: prometheus_rules_file is defined and prometheus_rules_file == 'default' and prom_rules_sub_file.stat.exists
+  when: (prometheus_rules_file is defined) and (prometheus_rules_file != 'default') and (prom_rules_sub_file.stat.exists)
```
vladzcloudius, line 121: And what if `prometheus_rules_file` is not defined? Don't we want to copy ALL rules files regardless?

vladzcloudius (changes requested):
This PR seems to be breaking the support for Monitoring before 4.x. 

There were other comments too.

In general a lot more detailed description of the patch is expected: why this change is needed and what this PR is going to change to make it better?

ansible-scylla-monitoring/tasks/common.yml
tarzanek: nope, this is only in case you want to amend/add new rules not present in old monitoring
HOWEVER it breaks old behaviour, where you could completely wipe out what monitoring provides by your own config file (do we even want to support that?)

ansible-scylla-monitoring/tasks/common.yml
tarzanek: fixed, bug on my side, found it in testing now

tarzanek:
retest with smallest supported version

tarzanek:
big question is if we can disable default alerts from prometheus default by overriding it with our custom file
or on other level (e.g. alertmanager)

tarzanek:
@vladzcloudius how about we split the need to add alert disable and implement it clearly and not as part of this PR?
We don't use alert disable anywhere and previous way was bad anyhow.
So let's make this PR as fix for monitoring (which basically needs this change since 3.8) and let's add separate feature to disable alerts as some other PR, how does that sound?

PR #236, opened 2023-03-27
ansible-scylla-node: Use already existing io_properties.yaml by default

igorribeiroduarte:
Currently, by default, the role will always try to replace the io_properties.yaml file according to scylla_io_setup output or pre-defined io_properties variable. However, it makes more sense to avoid touching this file if it already exists, since the user might be, for example, performing some kind of test which requires io_properties to use different values for different nodes.
This patch changes the default behavior, and now io_properties.yaml will only be replaced if the flag `always_replace_io_properties`, also added on this patch, is set to True.


vladzcloudius (approved):

vladzcloudius (approved):
Looks good. 
Igor, could you make sure CI passes with this and I'll merge it.

PR #240, opened 2023-04-05
ansible-scylla-monitoring: Move scylla_servers.yml generation to the monitoring role

igorribeiroduarte:
Currently, scylla_servers generation is done in the node role and this creates an unnecessary dependency from the monitoring role on the node role, since this file is useful only for the former.
This patch fixes this by moving the files related to the generation of this file to the monitoring role. It also renames monitoring_config.yml to generate_scylla_servers.yml.


tarzanek (approved):
lgtm

ansible-scylla-monitoring/tasks/generate_scylla_servers.yml
```diff
@@ -5,10 +5,11 @@
   args:
```
tarzanek, line 5: {{ scylla_api_address }}:{{ scylla_api_port }}

is also not part of monitoring role ...

tarzanek: IF we are going to move to use https://github.com/scylladb/scylla-monitoring/blob/master/genconfig.py
then those params are likely not needed and inventory can be used, also minigenconfig.py will be not needed (it is buggy anyways)

tarzanek: above said ... it's script Glauber wrote a LONG time ago, so it would be good to review it, it should work, but we should be sure and ideally @amnonh should bless its usage

tarzanek: AH, they are hardcoded above ... OK ... likely not what we want

igorribeiroduarte: I replaced the usage of `minigenconfig.py` with `genconfig.py`, as requested

tarzanek (approved):
lovely, more code removed than added!
kudoz! :-)

tarzanek:
@ebenzecri please if you have comments let them still be written down here so @igorribeiroduarte can address
thanks guys!

ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -73,11 +73,30 @@
...
     src: "{{ scylla_servers_yml_path }}"
     dest: "{{ scylla_monitoring_config_path }}/scylla_servers.yml"
     mode: '0644'
+  when: scylla_servers_stat.stat.exists|bool
+
+- name: generate scylla_servers.yml for monitoring
+  block:
+    - command: nodetool status
+      register: nodetool_status_out
+      delegate_to: "{{ groups['scylla'][0] }}"
```
tarzanek, line 93: actually one improvement for future would be to pickup a random node (or if you provide an index, then index of that node, since if above node is down, monitoring role won't get the status)

PR #241, opened 2023-04-06
manage_user.yml: Use import_task instead of include_tasks for creating admin_user

igorribeiroduarte:
Passing ignore errors to the task included works only with import_tasks. This patch replaces include_tasks with import_tasks in order for the ignore_errors to work. This ignore_errors is necessary because it's ok for the task of creating scylla_admin to fail if scylla_admin already exists.


tarzanek:
we should check if it exists and skip this task

tarzanek:
merging, but I expect a test for user so we can skip adding it
or recognize the error when the user is there using ansible

PR #242, opened 2023-04-13
ansible-scylla-monitoring: Allow gathering names/ips from the inventory to generate scylla_servers.yml

igorribeiroduarte:
Currently we're using the information from nodetool status to generate scylla_servers.yml.
However, it might be useful to use the information from the inventory, if we want, for example,
to use the dns names. This patch adds a variable `use_nodetool_status_with_genconfig` and sets it
to False as default. Whenever this variable is set to False, the role will generate scylla_servers.yml
based on the information from the inventory. Note that in order for this to work, the dc variable
must be defined for all the nodes in the inventory.
e.g.:
```
[scylla]
node1 dc=dc1
node2 dc=dc1
node3 dc=dc1
```


tarzanek:
who does the resolving?
or we assume inventory always has hostnames?

PR #243, opened 2023-04-25
Fix check mode in Debian installation

micwiel:


tarzanek:
what problem does this PR solve @micwiel ?

micwiel:
Please run the role with `--check` mode enabled, many tasks would fail but these problems could be easily handled by improvements i've implemented. They don't change anything in the role logic/design and should be completely safe.

PR #245, opened 2023-06-13
ansible-scylla-node: Automatically adjust replication for system_distributed and system_traces keyspaces

igorribeiroduarte:


vladzcloudius:
@igorribeiroduarte Now we need to instrument the Job to set those to `True` by default and have a way to override it.

cc @tarzanek

PR #253, opened 2023-06-21
ansible-scylla-node: Improves "legacy" support for 'node-exporter'

### LINKED ISSUE #249: ansible-scylla-node: 'scylla-node-exporter' service is not enabled by default
'scylla-node-exporter' service is started but not enabled automatically by the role.

ebenzecri:
This patch improves how 'node_exporter_install' is called and how 'node-exporter' service is started.

Fixes: #249

### DIFF
--- a/ansible-scylla-node/handlers/main.yml
+++ b/ansible-scylla-node/handlers/main.yml
@@ -24,12 +24,38 @@
     name: scylla-server
     state: stopped
 
-- name: node_exporter start
+- name: Enable and start 'node-exporter.service' service (legacy)
+  block:
+  - name: Disable and stop 'scylla-node-exporter.service' service
+    ansible.builtin.service:
+      name: scylla-node-exporter.service
+      state: stopped
+      enabled: no
+    when: ansible_facts.services["scylla-node-exporter.service"] is defined
+
+  - name: Enable and start 'node-exporter.service' service (legacy)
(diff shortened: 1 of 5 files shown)


ebenzecri:
### Installing Scylla Enterprise 2021 from scratch

- `node_exporter_install` is executed
- `node-exporter.service` is enabled and started
- `scylla-node-exporter.service` is not present
- Molecule tests passed

### Installing Scylla Enterprise 2022 from scratch
- `node_exporter_install` is not executed
- `node-expoter.service` is not present
- `scylla-node-exporter.service` is enabled and started
- Molecule tests passed

### Upgrading Scylla Enterprise from 2021 to 2022
Here I've spotted a problem related to systemd units. After upgrading 2022.2, the unit file corresponding to `node-exporter.service` is removed. Let me show an example if `systemctl status node-exporter.service` is executed:
```
● node-exporter.service
    Loaded: not-found (Reason: Unit node-exporter.service not found.)
    Active: active (running) since Wed 2023-06-21 17:20:46 UTC; 1h 39min ago
```
So, if I run `systemctl start node-exporter.service`, it will fail.

The solution here seems to be calling a handler for reloading systemd configuration (there's an unused one there), but probably Molecule will fail (because Scylla images aren't using systemd). All handlers that manages systemd unit must be skipped from testing.

PR #260, opened 2023-07-10
Refactor replace playbook

igorribeiroduarte:


ansible-scylla-node/tasks/common.yml
```diff
@@ -67,7 +67,7 @@
 
     - set_fact:
         io_conf: "{{ _io_conf.stdout_lines[0] }}"
-  when: scylla_io_probe|bool == True and (scylla_io_probe_dc_aware|bool == False or dc_to_node_list[dc][0] == inventory_hostname)
+  when: io_properties is not defined and scylla_io_probe|bool == True and (scylla_io_probe_dc_aware|bool == False or dc_to_node_list[dc][0] == inventory_hostname)
```
vladzcloudius, line 70: Your usage of `io_settings` term in the patch description is confusing.
As far as I know there is no such term.

You probably wanted to write "I/O settings" instead.

vladzcloudius (changes requested):

ansible-scylla-node/templates/scylla.yaml.j2
```diff
@@ -21,7 +21,11 @@ commitlog_segment_size_in_mb: 32
 seed_provider:
     - class_name: org.apache.cassandra.locator.SimpleSeedProvider
       parameters:
+{% if temporary_seed is defined %}
```
vladzcloudius, line 24: Why do you need this variable?
Why not to auto-select a seeds for a node being replaced as follows:

Names:
* Node A - node being replaced
* Node B - node that replaced Node A

1) Use the first node from the inventory which is not Node B (when you run this playbook inventory is supposed not to have Node A anymore) as a seeder for Node B.
2) After Node B bootstraps change its seeder list as follows:
  * If the original seeders list included "Node A" - replace Node A with Node B and use that list. 
  * Otherwise use seeders list as-is.
3) On every other node replace Node A with Node B in seeders list.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/common.yml
```diff
@@ -285,6 +285,11 @@
   set_fact:
     broadcast_address: "{{ scylla_broadcast_address }}"
 
+- name: Set start_scylla_service to false if there is an ongoing replacement
+  set_fact:
+    _start_scylla_service: |
+      {% if (start_scylla_service and replacing_node == false) %}True{% else %}False{% endif %}
```
vladzcloudius, line 291: That's a little clumsy expression. 
Why not simply:

```
- name: Set start_scylla_service to false if there is an ongoing replacement
  set_fact:
    _start_scylla_service: '{{ start_scylla_service == true and replacing_node == false }}'
```
??

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,66 @@
...
-    # Use the same parameters as for the rest of the cluster with the addition of the following:
-    scylla_yaml_params:
-      replace_address_first_boot: {{ replaced_node }}
-
+# This play will make sure that the node being replaced is down and prevent it from trying to re-join
+# the cluster by blacklisting its ip via iptables. It'll also define a temporary_seed var which will
+# later be used by the other plays.
+- name: Prepare nodes before replacement
+  hosts: scylla
+  become: false
```
vladzcloudius, line 9: that's a little unexpected...

ansible-scylla-node/templates/scylla.yaml.j2
vladzcloudius: I see that you kinda doing the above in following patches in the external playbook.
What I don't get why you need an additional variable here and can't simply use `scylla_seeds`?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,66 @@
...
 
-  tasks:
-    - name: start scylla
-      service:
-        name: scylla-server
-        state: started
-      become: true
+    - name: Check if CQL port is up on {{ replaced_node_broadcast_address }}
+      wait_for:
+        port: 9042
```
vladzcloudius, line 28: Let's parametrize the CQL port, please.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
+        regexp: '^replace_node_first_boot:'
+        state: absent
+      when: _replace_node_first_boot_grep.failed == false
+
+    - name: Remove the replace_address_first_boot record
+      lineinfile:
+        path: /etc/scylla/scylla.yaml
+        regexp: '^replace_address_first_boot:'
+        state: absent
+      when: _replace_node_first_boot_grep.failed
```
vladzcloudius, line 175: Don't you need this to be `_replace_node_first_boot_grep.failed|bool` or `_replace_node_first_boot_grep.failed == true` ?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
+        state: restarted
+        enabled: yes
+
+    - name: Check if RBNO is available in the current Scylla version
+      shell: |
+        scylla --help | grep enable-repair-based-node-ops
+      register: _enable_rbno_grep
+      ignore_errors: true
+
+    - name: Check if RBNO is enabled for node replacement
```
vladzcloudius, line 189: At this stage it's "if RBNO was enabled" ;)

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
+        - command: cat /etc/scylla/scylla.yaml
+          ignore_errors: true
+          register: _scylla_yaml_out
+
+        - set_fact:
+            _scylla_yaml_map: "{{ _scylla_yaml_out.stdout | from_yaml }}"
+
+        - set_fact:
+            _rbno_enabled: |
+              {% if (_scylla_yaml_map.enable_repair_based_node_ops is not defined or _scylla_yaml_map.enable_repair_based_node_ops == true) %}True{% else %}False{% endif %}
```
vladzcloudius, line 200: Here you can also simplify this expression

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
+            _rbno_allowed_for_replace: |
+              {% if (_scylla_yaml_map.allowed_repair_based_node_ops is not defined or 'replace' in _scylla_yaml_map.allowed_repair_based_node_ops) %}True{% else %}False{% endif %}
```
vladzcloudius, line 202: ... and here

example-playbooks/replace_node/replace_node.yml
```diff
@@ -195,3 +195,39 @@
...
+
+
+# This play will repair the new node if RBNO was not used and skip_repair is not set to true
+- name: Repair the new node, if necessary
+  hosts: scylla-manager
+  become: false
+  tasks:
+    - name: Skip this playbook if RBNO was used for the replacement or if skip_repair is true
+      meta: end_host
+      when: skip_repair|bool or (hostvars[new_node]['_enable_rbno_grep'].failed == false and hostvars[new_node]['_rbno_enabled'] and hostvars[new_node]['_rbno_allowed_for_replace'])
```
vladzcloudius, line 215: Don't you need to use `|bool` for every boolean you check to be `True` in this expression?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -195,3 +195,39 @@
...
       when: _enable_rbno_grep.failed == false
+
+
+# This play will repair the new node if RBNO was not used and skip_repair is not set to true
+- name: Repair the new node, if necessary
+  hosts: scylla-manager
+  become: false
+  tasks:
+    - name: Skip this playbook if RBNO was used for the replacement or if skip_repair is true
+      meta: end_host
```
vladzcloudius, line 214: `end_play` would be more readable here.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/common.yml
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
+      shell: |
+        nodetool status|grep -E '^UN|^UJ|^DN'|grep "{{ vars['ansible_'~scylla_nic].ipv4.address }}"| wc -l
+      register: node_count
+      until: node_count.stdout|int == 1
+      retries: 300
+      delay: 1
+
+    - name: Make sure that the seed also sees the new node in the UN state
+      shell: |
+        nodetool status|grep -E '^UN|^UJ|^DN'|grep "{{ vars['ansible_'~scylla_nic].ipv4.address }}"| wc -l
```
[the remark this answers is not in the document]
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,66 @@
...
+      loop: "{{ ansible_play_hosts_all }}"
+      run_once: true
 
-  tasks:
-    - name: start scylla
-      service:
-        name: scylla-server
-        state: started
-      become: true
+    - name: Check if CQL port is up on {{ replaced_node_broadcast_address }}
```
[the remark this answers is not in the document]
igorribeiroduarte: It's a mandatory parameter. It's listed in `vars/main.yml`

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: No, we only need to cast if we're not sure that the variable is boolean.
For example, for the variable `skip_repair` the user could pass `skip_repair: 'false'` instead of `skip_repair: false`, so it would make sense to cast the variable to boolean.
But in this specific case we know that `failed` is always a boolean.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: No, we only need to cast if we're not sure that the variable is boolean.
For example, for the variable `skip_repair` the user could pass `skip_repair: 'false'` instead of `skip_repair: false`, so it would make sense to cast the variable to boolean.
But in this specific case we know that `failed`, `_rbno_enabled` and `_rbno_allowed_for_replace` are always a boolean.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Initially I was using `become: true` for all the plays, but I had problems because some tasks from the node role will fail if you try to run them with `sudo`.
So I noticed that in the node role the tasks which need to be executed with `sudo` have `become: true` already, so by setting `become: false` in the play I make sure that only the tasks which really need `sudo` will be executed with `sudo`

In this specific play (prepare nodes before replacement), the only task which needs `become: true` is the one using `iptables`

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
       delegate_to: "{{ temporary_seed }}"
       run_once: true
+
+
+# This play will install Scylla in the new node using the temporary_seed var (defined in the previous play)
+# as a seed. The same variables used when the node role was executed for the replaced node should also
+# be passed to this playbook and will be used by this play.
+- name: Install Scylla in the new node
+  hosts: "{{ new_node }}"
+  become: false
```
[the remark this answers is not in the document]
igorribeiroduarte: Same as https://github.com/scylladb/scylla-ansible-roles/pull/260#discussion_r1261157396

example-playbooks/replace_node/replace_node.yml
```diff
@@ -64,3 +64,134 @@
...
+    - name: Start Scylla
+      service:
+        name: scylla-server
+        state: started
+
+    - name: Wait for CQL port on {{ listen_address }}
+      wait_for:
+        port: 9042
+        host: "{{ listen_address }}"
+        timeout: "{{ scylla_bootstrap_wait_time_sec }}"
```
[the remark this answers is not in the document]
igorribeiroduarte: This variable is from the node role and I shouldn't be using it here.
I replaced it with `new_node_bootstrap_wait_time_sec` and added this variable to `vars/main.yml`

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,66 @@
...
-    - name: check for the replaced node being a seed
-      fail:
-        msg: "The node being replaced {{ replaced_node }} is a seed node. Please demote it to a non-seed first and try again."
-      when: replaced_node in scylla_seeds
-
-  roles:
-    - ansible-scylla-node
+    - name: Resolve a scylla_listen_address as a fact
+      set_fact:
+        listen_address: "{{ scylla_listen_address }}"
```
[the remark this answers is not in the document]
igorribeiroduarte: This variable is from the node role and I shouldn't be using it here.
I replaced both `scylla_listen_address` and `scylla_broadcast_address` with `alive_nodes_listen_address` and `alive_nodes_broadcast_address` and added these variables to `vars/main.yml`

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

vladzcloudius:
@igorribeiroduarte I still don't understand why do we need all this extra boilerplate of adding these new additional Role variables instead of invoking the existing Role with parameters we need from some external Python or even Ansible using `include_role`?

All those new variables IMO create unnecessary noise and dependencies we don't want to have.

vladzcloudius (changes requested):

vladzcloudius:
@ebenzecri FYI.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
-  roles:
-    - ansible-scylla-node
+    - name: Resolve scylla_broadcast_address as a fact
+      set_fact:
+        broadcast_address: "{{ alive_nodes_broadcast_address }}"
 
+    - fail:
+        msg: "The replaced node can't be a seed!"
+      when: replaced_node_broadcast_address in scylla_seeds
+      run_once: true
```
vladzcloudius, line 20: What if it is?
Seeds can die too.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
 
-- name: Start scylla services
-  # new_node should be the IP of the node that comes into the cluster. The node being replaced should be in `replaced_node` variable below
-  hosts: new_node
-  become: true
+    - fail:
+        msg: "The new node can't be a seed!"
+      when: new_node in scylla_seeds
+      run_once: true
```
vladzcloudius, line 25: What if it is replacing a dead seed?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
-      service:
-        name: scylla-server
-        state: started
-      become: true
-      ignore_errors: true
-      #TODO: remove ignore_errors when ansible is bumped to 2.10.4 or 2.9.16 as per https://github.com/ansible/ansible/issues/71528
+    - name: Save one of the seeds as a fact in order to run some checks
+      set_fact:
+        alive_node: "{{ item }}"
+      when: item != replaced_node and item != new_node
```
vladzcloudius, line 30: You have "fail" blocks that validate that these conditions are never met.
I'm confused.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
+- name: Prepare nodes before replacement
+  hosts: scylla
   tasks:
-    - name: check for the replaced node being a seed
-      fail:
-        msg: "The node being replaced {{ replaced_node }} is a seed node. Please demote it to a non-seed first and try again."
-      when: replaced_node in scylla_seeds
+    - name: Resolve a scylla_listen_address as a fact
+      set_fact:
+        listen_address: "{{ alive_nodes_listen_address }}"
```
vladzcloudius, line 11: Where does `alive_nodes_listen_address` come from?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
 
-  roles:
-    - ansible-scylla-node
+    - name: Resolve scylla_broadcast_address as a fact
+      set_fact:
+        broadcast_address: "{{ alive_nodes_broadcast_address }}"
```
vladzcloudius, line 15: Same here: where does `alive_nodes_broadcast_address` come from?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
-        nodetool status|grep -E '^UN|^UJ|^DN'|grep "{{ vars['ansible_'~scylla_nic].ipv4.address }}"| wc -l
-      register: node_count
-      until: node_count.stdout|int == 1
-      retries: 300
-      delay: 1
+    - name: Check if CQL port is up on {{ replaced_node_broadcast_address }}
+      wait_for:
+        port: "{{ cql_port }}"
+        host: "{{ replaced_node_broadcast_address }}"
+        timeout: 60
```
vladzcloudius, line 38: Parametrise the timeout, please.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
 
-    - name: run nodetool repair -pr
-      shell: |
-        nodetool repair -pr
+    - name: Check if the other nodes are UP
+      wait_for:
+        port: "{{ cql_port }}"
+        host: "{{ listen_address }}"
+        timeout: 60
+      when: inventory_hostname != replaced_node and inventory_hostname != new_node
```
vladzcloudius, line 55: We had to handle the multiple nodes replacement with @ebenzecri yesterday.
I hope this is going to be the last time we see a shit-show like that but this proves that this is quite possible.

If we implement this playbook to require up to a single node DOWN we can hack multiple-nodes-DOWN around by commenting "other DOWN nodes" in the inventory while we run this.

However I wonder how much more complexity supporting multiple dead nodes would add? 

WDYT, @igorribeiroduarte ?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -1,64 +1,74 @@
...
+      when: _wait_for_cql_port_output.failed|bool == false
+      run_once: true
 
-    - name: run nodetool repair -pr
-      shell: |
-        nodetool repair -pr
+    - name: Check if the other nodes are UP
+      wait_for:
+        port: "{{ cql_port }}"
+        host: "{{ listen_address }}"
```
vladzcloudius, line 53: Shouldn't you be checking `broadcast_rpc_address`?
For example, on AWS with a public IP configured and used for CQL `listen_address` is going to be an internal IP while clients (and this playbook) should connect to a CQL port via the public IP. 
And this is configured via `broadcast_rpc_address` parameter.

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
```diff
@@ -72,3 +72,129 @@
...
+
+# This play will install Scylla in the new node and update the existing nodes.
+# The same variables used when the node role was executed for the replaced node should also
+# be passed to this playbook and will be used by this play.
+- name: Install/Update Scylla in all nodes
+  hosts: scylla
+  vars:
+    firewall_enabled: true
+  roles:
+    - ansible-scylla-node
```
vladzcloudius, line 89: Don't you want to set `don't start scylla service` somewhere? 
If you did, where is it?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -198,3 +198,38 @@
   hosts: scylla-monitor
   roles:
     - ansible-scylla-monitoring
+
```
vladzcloudius, line 207: In the patch description: s/Manually repairing/Repairing/
"Manual" means that actual person has to do that.

vladzcloudius (changes requested):

example-playbooks/replace_node/vars/main.yml
```diff
@@ -0,0 +1,36 @@
...
+
+# Any extra params you might want to pass sctool repair, like --intensity and/or --parallel
+extra_repair_params: ''
+
+# The maximum time we'll wait for the repair to finish before
+# stopping the execution of the playbook with an error message.
+new_node_repair_timeout_seconds: 36000
+
+# Default bootstrap timeout for the new node
+new_node_bootstrap_wait_time_sec: 25200
```
vladzcloudius, line 31: Here, you already parametrized it! :D
Let's use it!

example-playbooks/replace_node/vars/main.yml
```diff
@@ -0,0 +1,36 @@
...
+
+alive_nodes_listen_address: "{{ vars['ansible_'~scylla_nic].ipv4.address }}"
+alive_nodes_broadcast_address: "{{ vars['ansible_'~scylla_nic].ipv4.address }}"
```
vladzcloudius, line 34: Names of these are a bit confusing. Aren't these simply `listen_address` and `broadcast_address`?

example-playbooks/replace_node/replace_node.yml
vladzcloudius: I know the answer now. 
But you probably want to rebase and change the order of your patches to have a patch with `main.yml` before the one that uses variables defined by it.

example-playbooks/replace_node/replace_node.yml
vladzcloudius: You seem to forget to use the `new_node_bootstrap_wait_time_sec` here.

example-playbooks/replace_node/vars/main.yml
igorribeiroduarte: I used these names to make it clear that these are the addresses of the alive nodes to avoid confusion with the replaced node, since we have the variable `replaced_node` and `replaced_node_broadcast_address` for it.

And we need `replaced_node` addresses to be separate variables because the replaced node is not part of the inventory when this playbook is called.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: It's set in `vars/main.yml`

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: These tasks run in the nodes themselves. So for node A this task will be executed in node A, and for node B it'll be executed in node B, so I think it makes sense to use `listen_address` here, right?

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: > However I wonder how much more complexity supporting multiple dead nodes would add?

Not much. If I'm not missing anything, we'd only have to:
* Ajdust the `Check if other nodes are UP` and the `Validate that the node being replaced is down`  tasks
* Make the `new_node` var a list
* Change `replaced_node` and `replaced_node_broadcast_address`  to lists

However, I'd suggest we do this as a follow-up PR since this would require testing everything again and it would delay the review.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: In this task I'm not bootstrapping the new node.
I'm using this only to validate that the replaced node is down and all the others are up before I start replacing the node, so
a fixed small timeout should be enough IMO.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: My idea is that the user calling this playbook will already pass an updated seeds list.
So this shouldn't be a problem.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: My idea is that the user calling this playbook will already pass an updated seeds list.
So this shouldn't be a problem.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

example-playbooks/replace_node/vars/main.yml
vladzcloudius: But wouldn't these include the value for `replaced_node` too?

vladzcloudius (changes requested):
Still have one comment.

example-playbooks/replace_node/README.md
```diff
@@ -2,40 +2,50 @@
...
 
-### Steps:
+## Prerequisites
 
-1. Check if the node is a seed, if it is stop, seeds need to be demoted first.
-2. Apply the ansible-scylla-node role to the new node, reusing all the parameters from the existing cluster. Skip starting the scylla-server service
-3. Start the services
-4. Clean out the replace_address_first_boot record from scylla.yaml
-5. Run `nodetool repair -pr`
+* A scylla cluster with scylla-manager and scylla-monitoring installed.
```
vladzcloudius, line 7: Here, I think you should add a section about the inventory content before you run this playbook. Am I right?

You could say something like: "Make sure to update the inventory file as described in the Usage section below before running the playbook."

example-playbooks/replace_node/README.md
```diff
@@ -2,40 +2,50 @@
...
+In the `Usage` section of this README, we use `nodes_params.yml` and `monitor_params.yml` files to represent such parameters.
+
+Besides the vars from the node role, this playbook has the following mandatory parameters: `replaced_node`, `replaced_node_broadcast_address` and `new_node`.
+These and the other available parameters are listed and described in `vars/main.yml`.
 
 ## Usage:
 
-1. Reuse the existing set of parameters from the ansble-scylla-node run for the existing cluster. In this case I copy a previously created file called `nodes_params.yml` to `replace_params.yml`.
-2. Add the new node to the inventory:
+1. Add the new node to the inventory:
```
vladzcloudius, line 33: This step in not optional. Hence you want to either mention it explicitly or reference it from the "Prerequisites" section.

example-playbooks/replace_node/README.md
```diff
@@ -2,40 +2,50 @@
...
+## Parameters:
+
+This playbook uses the node role to install and configure Scylla in the new node, so the same parameters
+used when the cluster was created should also be passed to the `replace_node.yml` playbook.
+In the `Usage` section of this README, we use `nodes_params.yml` and `monitor_params.yml` files to represent such parameters.
+
+Besides the vars from the node role, this playbook has the following mandatory parameters: `replaced_node`, `replaced_node_broadcast_address` and `new_node`.
+These and the other available parameters are listed and described in `vars/main.yml`.
 
 ## Usage:
```
vladzcloudius, line 31: Please, move this section to the top of the file. Please, put it right after the "Prerequisites" section.

These two sections are ones that 99% of the users of this playbook are going to be reading/looking for.

vladzcloudius (changes requested):
Looks good except for some comments about the README.md.

Let's fix them and it's good to be merged.

example-playbooks/replace_node/README.md
```diff
@@ -2,40 +2,50 @@
...
-3. Start the services
-4. Clean out the replace_address_first_boot record from scylla.yaml
-5. Run `nodetool repair -pr`
+* A scylla cluster with scylla-manager and scylla-monitoring installed.
+
+## Steps:
+
+1. Validate that the node being replaced is down
+2. Block the broadcast address of the replaced node via iptables in all the other nodes to prevent it from trying to join the cluster again
+3. Check if all the other nodes are up
```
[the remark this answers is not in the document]
igorribeiroduarte: the way it is right now, any node different from the replaced node and from the new node is considered an alive node, and by default their addresses are the ipv4 addresses of the default nic of each node.

example-playbooks/replace_node/README.md
```diff
@@ -2,40 +2,50 @@
...
+9. Update Scylla-Monitoring
+10. If RBNO is disabled, repair the new node
+
+## Parameters:
+
+This playbook uses the node role to install and configure Scylla in the new node, so the same parameters
+used when the cluster was created should also be passed to the `replace_node.yml` playbook.
+In the `Usage` section of this README, we use `nodes_params.yml` and `monitor_params.yml` files to represent such parameters.
+
+Besides the vars from the node role, this playbook has the following mandatory parameters: `replaced_node`, `replaced_node_broadcast_address` and `new_node`.
```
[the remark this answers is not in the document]
igorribeiroduarte: They are not mandatory, since we have the ipv4 address of the default nic as a default, in the same way we do with the node role.

example-playbooks/replace_node/README.md
igorribeiroduarte: Done

example-playbooks/replace_node/README.md
igorribeiroduarte: Done

example-playbooks/replace_node/README.md
igorribeiroduarte: Done

PR #266, opened 2023-07-22
scylla-monitoring: rule_config.yml.j2: add routes for string severities

vladzcloudius:
Since "prometheus.rules.yml: Use severity as string instead of numbers" Alert Manager is going to use string severities values, like "warn", "error", etc.

However the same patch introduced a regression by inverting the severities: https://github.com/scylladb/scylla-monitoring/issues/2029

While we are waiting for a fix let's "hack" it so that things continue working.


AdamStawarz (approved):
reviewed

d-helios (approved):

PR #278, opened 2023-08-05
scylla-node: scylla_raid_setup: allow overriding "online discard" option

vladzcloudius:
This is a dangerous option that should not be used by default.

This allows specifying whether one wants it to be used or not. By default we will not use it.

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -303,6 +303,7 @@ scylla_manager_agent_config: |
   # - /dev/nvme0n1
   # - /dev/nvme0n2
   # - /dev/nvme0n3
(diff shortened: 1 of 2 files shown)


tarzanek:
lgtm, can we merge or is this waiting for https://github.com/scylladb/scylladb/pull/14964 ?
(I hope not, but I want to make sure) @vladzcloudius ?

vladzcloudius:
> lgtm, can we merge or is this waiting for [scylladb/scylladb#14964](https://github.com/scylladb/scylladb/pull/14964) ? (I hope not, but I want to make sure) @vladzcloudius ?

Thanks.
No, we don't need to wait. It just won't do till the fix in `scylla` is merged.
We will use the patched version as long as needed.

PR #288, opened 2023-09-22
scylla-node: don't upgrade already installed scylla unless explicitly requested

vladzcloudius:


igorribeiroduarte (approved):

vladzcloudius:
Guys, please, don't merge yet. I'll do it myself.

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -74,12 +74,15 @@
       name: java
       path: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
 
-  - name: "Going to install {{ scylla_version }}"
-    set_fact:
-      scylla_version_to_install: "{{ scylla_version }}"
-
-  - name: Install Scylla packages
-    include_tasks: Debian_install.yml
```
tarzanek, line 82: introduce instead a check if versions match or not and stop before doing damage

ansible-scylla-node/defaults/main.yml
```diff
@@ -200,6 +200,12 @@ scylla_version: '0.0.0'
 #scylla_edition: oss
 scylla_edition: enterprise
 
+# True if Scylla is already installed
+scylla_is_installed: False
+
+# Set to true if you want to allow a Role to upgrade and already installed Scylla
+scylla_allow_upgrades: False
```
tarzanek, line 207: rename to scylla force package install instead?

upgrade is elsewhere

tarzanek:
let's write the proper version to metadata/inventory to error out

vladzcloudius:
We brainstormed with Lubos and it was decided to rewrite this PR completely.
The idea is to implement the following logic:

1) The flow in question will **never** upgrade/downgrade the already installed scylla version.
2) The Role by default will error out if the installed scylla version is different from the one specified in `scylla_version`. 
3) The Role will always install the scylla version specified in `scylla_version` on the node that doesn't have any scylla installed.
4) There will be a flag that would allow disabling the check (2).

PR #289, opened 2023-09-23
scylla-node: don't ever install scylla on top of already installed one

vladzcloudius:
Scylla should only be upgraded via the corresponding 'upgrade' flow. Other than that the Role should only install binaries if they haven't been installed yet.

If they have already been installed but the version differs from the one specified via 'scylla_version' Role will error out by default.

The Operator may decide to skip the check above using 'skip_installed_scylla_version_check'.

### DIFF
--- a/ansible-scylla-node/tasks/Debian_install.yml
+++ b/ansible-scylla-node/tasks/Debian_install.yml
@@ -21,6 +21,11 @@
         msg: "Too many/few choices for a requested version '{{ scylla_version_to_install }}': {{ aptversions.stdout_lines }}. Bailing out!"
       when: aptversions.stdout_lines | length != 1
 
+    - name: If Scylla is installed check if it's the same version as requested
+      fail:
+        msg: "Installed version ({{ installed_scylla_version }}) doesn't match a requested one ({{ aptversions.stdout }})"
+      when: scylla_is_installed and not skip_installed_scylla_version_check and installed_scylla_version != aptversions.stdout
+
     - name: Do install Scylla
       block:
         - name: Nuke a {{ scylla_edition }} pin file if exists
@@ -48,4 +53,5 @@
             name: "{{ scylla_package_prefix }}={{ aptversions.stdout }}"
             state: present
             allow_downgrade: yes
+      when: not scylla_is_installed
   become: true
--- a/ansible-scylla-node/tasks/Debian_install.yml
+++ b/ansible-scylla-node/tasks/Debian_install.yml
@@ -46,8 +46,9 @@
         # 'apt list -a' output has a package version as a second column and an arch as a third one.
         # Let's filter by the arch first and then cut the version column.
         # Then we will filter out all rows that start with a requested version string followed by a digit to filter out version like 2021.1.11 when 2021.1.1 was requested.
-        # And finally, we are going to get rid of duplications.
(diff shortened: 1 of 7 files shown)


ansible-scylla-node/tasks/main.yml
```diff
@@ -8,6 +8,51 @@
...
+  set_fact:
+    scylla_version_to_install: "{{ scylla_version }}"
+
+- name: Get latest Scylla {{ scylla_package_prefix }} version if needed
+  block:
+    - name: Get OSS latest version
+      ansible.builtin.uri:
+        url: "https://repositories.scylladb.com/scylla/check_version"
+        method: GET
+        register: scylla_latest_version
```
tarzanek, line 41: so there can be mismatch in https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-node/requirements.txt
and ansible version 6.7
""Unsupported parameters for (ansible.legacy.uri) module: register."

but more likely tests fail due to extra TAB before register - it should be on the same level as "name"
so please fix that(just move this line few spaces back to "- name" level)

ansible-scylla-node/tasks/main.yml
```diff
@@ -8,6 +8,51 @@
...
+      when: scylla_edition == 'oss'
+
+    - name: Get Enterprise latest version
+      ansible.builtin.uri:
+        url: "https://repositories.scylladb.com/scylla/check_version?system=enterprise"
+        method: GET
+        register: scylla_latest_version
```
tarzanek, line 48: dtto as above

vladzcloudius:
New:
* Fixed things that made the molecule fail.
* Only fetch latest version when 'online' install type is used.
* Error out when 'latest' version is requested for a non-online installation.

ansible-scylla-node/tasks/main.yml
```diff
@@ -8,6 +8,52 @@
...
+    scylla_is_installed: "{{ true if scylla_package_prefix in ansible_facts.packages else false }}"
+
+- name: Get installed scylla version
+  set_fact:
+    installed_scylla_version: "{{ ansible_facts.packages[scylla_package_prefix][0]['version'] }}"
+  when: scylla_is_installed
+
+- name: Initialize scylla_version_to_install to {{ scylla_version }}
+  set_fact:
+    scylla_version_to_install: "{{ scylla_version }}"
```
tarzanek, line 32: won't this conflict with update job? ( I forgot the flow, but if this is inited before update job it will take over variable names in case they are the same here and there)

ansible-scylla-node/tasks/Debian_install.yml
```diff
@@ -1,84 +1,57 @@
...
-        set_fact:
-          scylla_package_prefix: "scylla"
-        when: scylla_edition == 'oss'
-
-      - name: Set Scylla package prefix as Enterprise
-        set_fact:
-          scylla_package_prefix: "scylla-enterprise"
-        when: scylla_edition == 'enterprise'
-
-      - name: Get versions of {{ scylla_edition }} package
```
tarzanek, line 45: it's hard to read the diffs here
was this just reorganization of blocks or also some rewrite/fixes in them?

ansible-scylla-node/tasks/main.yml
vladzcloudius: This fact is internal and if any other flow uses it it will overwrite it to whatever it needs.

ansible-scylla-node/tasks/Debian_install.yml
vladzcloudius: A lot of deleted pieces mostly since a lot of it moved to main.yaml.

Let me rebase the PR and split this patch into two.

vladzcloudius: Done.

PR #295, opened 2023-09-28
ansible-scylla-manager: Use provided scylla_version instead of hardcoded latest

igorribeiroduarte:


vladzcloudius:
The workaround is to provide a 2023.1 repo in addition to the one needed for provided scylla version.

vladzcloudius:
And no, we don't want to use a provided scylla version for a Manager Scylla.

PR #296, opened 2023-09-29
Scylla Node Role: implement repository auto-selection

vladzcloudius:


vladzcloudius:
Tested as follows:
1) Created a cluster of 2 nodes: one with Centos8 and another with Ubuntu 20.04
2) Used a Role from PR to deploy a cluster on a setup from (1) while not specifying any of `scylla_xx_repos` or `scylla_repos`: both RedHat and Ubuntu were successfully installed with a correct repo used.
Repeated (2) for `scylla_version: 2022.2.6` and `scylla_version: 'latest'`.
3) Tested that Scylla Manager Role is able to install 'latest' Scylla successfully.

tarzanek (approved):
lgtm!

PR #310, opened 2023-12-13
replace_node.yml: Allow replacing an alive node

igorribeiroduarte:


example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,30 @@
       when: hostvars[new_node]['broadcast_address'] in scylla_seeds or new_node in scylla_seeds
       run_once: true
 
+    - name: Stop scylla-server on the node being replaced
+      block:
+        - name: run nodetool drain
+          shell: |
+            nodetool drain
```
vladzcloudius, line 31: `nodetool drain` is blocking, so you would need to run it using `async_task`.
However why do you even drain this node? You are about to kill it!

example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,30 @@
...
+        - name: wait for DRAIN to complete
+          shell: |
+            journalctl -u scylla-server --since "1 minute ago" | grep 'storage_service' | grep 'DRAINED'
+          retries: 3000
+          delay: 1
+          register: drain_result
+          until: drain_result.rc == 0
+          ignore_errors: true
+
+        - name: Stop Scylla service
```
vladzcloudius, line 42: Continuing my previous point: you want to time limit this operation by, say 60 seconds and after that `kill -9` scylla. We really don't care about stop scylla "nice" on this node.

example-playbooks/replace_node/vars/main.yml
```diff
@@ -30,6 +30,10 @@ new_node_repair_timeout_seconds: 36000
 # Default bootstrap timeout for the new node
 new_node_bootstrap_wait_time_sec: 25200
 
+# If you set this to true, this playbook will allow you to replace an alive node
+# by draining/stopping the node being replaced before starting the replace
+alive_node_replace: false
```
vladzcloudius, line 35: I don't see a logic we discussed before: if `alive_node_replace == false` you wanted to check if we are not replacing an alive node my mistake: check if scylla service is running and if "yes" - fail angrily!

vladzcloudius (changes requested):

example-playbooks/replace_node/vars/main.yml
igorribeiroduarte: We were already doing that, so all I had to add was the task responsible for killing scylla if this flag is set to true.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,51 @@
...
+            enabled:  false
+          become: true
+
+        - name: Run nodetool drain
+          async_task:
+            shell: |
+              nodetool drain
+            alias: drain_scylla
+            async: 60
+            retries: 2
```
vladzcloudius, line 41: Why 2 retries?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,51 @@
...
+            delay: 30
+          ignore_errors: true
+          register: _drain_scylla_output
+
+        - name: nodetool drain logs
+          debug: var=_drain_scylla_output
```
vladzcloudius, line 47: nodetool drain doesn't print much AFAIR. Why is this needed?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,51 @@
...
+          ignore_errors: true
+          become: true
+
+        - name: Stop Scylla service
+          async_task:
+            shell: |
+              systemctl stop scylla-server
+            alias: stop_scylla
+            async: 60
+            retries: 2
```
vladzcloudius, line 61: Again, why 2 reties?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,51 @@
...
+        - name: nodetool drain logs
+          debug: var=_drain_scylla_output
+
+        - name: Send SIGKILL to scylla
+          shell: |
+            pkill -9 scylla$
+          ignore_errors: true
+          become: true
+
+        - name: Stop Scylla service
```
vladzcloudius, line 55: You needed to stop the service first and only then send SIGKILL. 
The way you do it here systemd is going to restart scylla-server service right after you kill it.

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Logs removed

example-playbooks/replace_node/replace_node.yml
```diff
@@ -24,6 +24,46 @@
...
+            delay: 1
+          ignore_errors: true
+          register: _drain_scylla_output
+
+        - name: Stop Scylla service
+          async_task:
+            shell: |
+              systemctl stop scylla-server
+            alias: stop_scylla
+            async: 60
```
vladzcloudius, line 51: We don't want to wait for the whole minute here. Here we want to wait, say 5 seconds and then kill the bastard (like you do below).

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

PR #313, opened 2023-12-15
molecule: Set force_schema_commit_log to true

igorribeiroduarte:
In the latest OSS version (5.4.0), if we use the defaults from Scylla we'll see the following error:
`init - Bad configuration: consistent_cluster_management requires schema commit log to be enabled`

This happens because consistent_cluster_management is set to true and force_schema_commit_log is set to false by default.

This PR fixes this error in our CI by setting force_schema_commit_log to true in scylla_yaml_params.


vladzcloudius (changes requested):
Please, reference a corresponding scylla GH issue - the defaults as described here are clearly broken.

igorribeiroduarte:
> Please, reference a corresponding scylla GH issue - the defaults as described here are clearly broken.

Done

PR #314, opened 2023-12-16
rule_config.yml.j2: include the alert name and severity in the subject

vladzcloudius:
Aggregate by an alert name and the severity (before that was only by severity) and then include both an alert name and its severity in the corresponding email subject.

As a result for tooManyFiles alert with a 'warn' severity a subject is going to look like this:

[PRODUCTION] tooManyFiles warn: URGENT Managed-Services - Cluster:cluster-name-goes-here

### DIFF
--- a/ansible-scylla-monitoring/templates/rule_config.yml.j2
+++ b/ansible-scylla-monitoring/templates/rule_config.yml.j2
@@ -50,7 +50,7 @@ receivers:
     auth_password: {{ alerts_sender_password }}
     send_resolved: false
     headers:
-      Subject: '[PRODUCTION] NORMAL Managed-Services - Cluster:{{ scylla_cluster_name }}'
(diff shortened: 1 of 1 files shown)


vladzcloudius:
cc @tomer-sandler

ansible-scylla-monitoring/templates/rule_config.yml.j2
```diff
@@ -61,9 +61,9 @@ receivers:
     auth_password: {{ alerts_sender_password }}
     send_resolved: false
     headers:
-      Subject: '[PRODUCTION] URGENT Managed-Services - Cluster:{{ scylla_cluster_name }}'
+      Subject: '[PRODUCTION] {{ '{{ .CommonLabels.alertname }}' }} {{ '{{ .CommonLabels.severity }}' }}: URGENT Managed-Services - Cluster:{{ scylla_cluster_name }}'
 route:
-  group_by: ['severity']
+  group_by: [alertname, 'severity']
```
igorribeiroduarte, line 66: Isn't it necessary for `alertname` to be between quotes?

igorribeiroduarte (approved):

PR #316, opened 2023-12-18
replace_node.yml: Adjust the task responsible for checking if the node is healthy

igorribeiroduarte:


example-playbooks/replace_node/replace_node.yml
```diff
@@ -155,7 +155,7 @@
 
     - name: Wait for the added node to become healthy
       shell: |
-        nodetool status|grep -E '^UN'|grep "{{ vars['ansible_'~scylla_nic].ipv4.address }}"| wc -l
+        nodetool status|grep -E '^UN'|grep " {{ vars['ansible_'~scylla_nic].ipv4.address }} "| wc -l
```
vladzcloudius, line 158: Why not using `grep -w ` instead?

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: I wasn't familiar with `grep -w`. It's indeed a better option.
Thanks.

PR #317, opened 2023-12-20
replace_node.yml: Use new_node_bootstrap_wait_time_sec as a timeout for 'Wait for the added node to becom healthy' task

igorribeiroduarte:


example-playbooks/replace_node/replace_node.yml
```diff
@@ -192,13 +192,14 @@
         port: "{{ cql_port }}"
         host: "{{ listen_address }}"
         timeout: "{{ new_node_bootstrap_wait_time_sec }}"
+      register: _wait_for_bootstrap
```
vladzcloudius, line 195: I don't see where it's used.

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: It's not. This was a leftover from another test I was doing.
It's removed now.

vladzcloudius (approved):

PR #318, opened 2023-12-26
replace_node.yml: Set silences in alertmanager before replacing the node

igorribeiroduarte:
This patch sets silences in 2 different situations:
  * In the node being replaced, when performing a live node replacement, since we know the node we're stopping will start to trigger alerts as soon as we stop it and these alerts will only stop once we update monitoring with the new node.
  * In the node being replaced and in the new node right after the 'Prepare nodes before replacement' play is done, since only after this play succeeds we know that we're actually good to start due to the validations performed in it.

This patch also expires the silences created once the replacement operations succeeds.

Note that if you're performing a dead node replacement and because of that this node was already triggering alerts, these alerts will continue until the 'Prepare nodes before replacement' play finishes, so you might still have to manually set a silence for the node being replaced.

### DIFF
--- a/example-playbooks/replace_node/replace_node.yml
+++ b/example-playbooks/replace_node/replace_node.yml
@@ -22,6 +22,15 @@
 
     - name: Stop scylla-server on the node being replaced
       block:
+        # Here we're setting a very small silence just to be able to finish checking if we should indeed move forward with this replace.
+        # Once we pass all the checks from the current play we'll set a new silence with a proper duration
+        - name: Set silence for the node that we are about to stop
+          shell: |
+            docker exec -it aalert amtool silence add instance="{{ replaced_node }}" --alertmanager.url="{{ alertmanager_url }}" --comment='Replacing node' --duration '5m'
+          # We don't want the playbook to fail in case we don't succeed in silencing the alerts
+          ignore_errors: true
+          delegate_to: "{{ groups['scylla-monitor'][0] }}"
+
         - name: Disable Scylla service
           service:
             name: scylla-server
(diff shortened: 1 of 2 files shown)


example-playbooks/replace_node/replace_node.yml
```diff
@@ -26,6 +26,13 @@
 
     - name: Stop scylla-server on the node being replaced
       block:
+        - name: Set silence for the node that we are about to stop
+          shell: |
+            docker exec -it aalert amtool silence add instance="{{ replaced_node }}" --alertmanager.url="{{ alertmanager_url }}" --comment='Replacing node' --duration "{{ new_node_bootstrap_wait_time_sec|int * 2 }}s"
+          # We don't want the playbook to fail in case we don't succeed in silencing the alerts
```
vladzcloudius, line 32: Why is that? What could be a good reason for such a failure?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -114,6 +121,20 @@
...
+- name: Silence alerts for the node being replaced
+  hosts: scylla-monitor
+  vars_files:
+    - vars/main.yml
+  tasks:
+    - name: Set silence for the node being replaced
+      shell: |
+        docker exec -it aalert amtool silence add instance="{{ new_node }}" --alertmanager.url="{{ alertmanager_url }}" --comment='Replacing node' --duration "{{ new_node_bootstrap_wait_time_sec|int * 2 }}s"
+      # We don't want the playbook to fail in case we don't succeed in silencing the alerts
+      ignore_errors: true
```
vladzcloudius, line 135: Same comment as above.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -127,12 +148,31 @@
     - ansible-scylla-node
 
 
+# Update scylla-monitoring before the replacement starts to be able to see the
+# streaming progress
```
vladzcloudius, line 152: Good comment but this change doesn't belong to this patch.
Send it as a separate patch, please.

example-playbooks/replace_node/replace_node.yml
```diff
@@ -127,12 +148,31 @@
...
   hosts: scylla-monitor
   roles:
     - ansible-scylla-monitoring
 
 
+# We want to silence alerts for the new node until the end of the replacement
+# The duration will be "new_node_bootstrap_wait_time_sec * 2", since this should
+# be enough for waiting for CQL port to become available and for waiting for the
+# node to become healthy.
+- name: Silence alerts for the new node
```
vladzcloudius, line 163: Does Monitoring restart clears previously set alerts silences? We do preserve the Prometheus "data". Doesn't it persists silences on the disk?

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: If monitoring is down or irresponsive for some reason, for example.

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Done

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Yes, it does persist silences.

PR #326, opened 2024-01-17
Node/Loader Roles: Debian.yml: remove explicit update-java-alternatives call

### LINKED ISSUE #325: [Node/Loader Roles]: Get rid of explicit arch mentioning of 'amd' arch in the code
HEAD: ff58a6035d03225ff9002714bc0a522372c38de9

*Description*
Two invocations of `alternatives` module refer an `amd` arch explicitly:

```
- name: Correct java version selected
    alternatives:
      name: java
      path: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
```

This makes this code not portable.

vladzcloudius:
There is no need to call update-java-alternatives explicitly since openjdk-8-jre-headless .deb package invokes it as a post-install action.

Tested on Ubuntu 20.04 on both x86 and on arm64 HW.

Fixes #325

### DIFF
--- a/ansible-scylla-loader/tasks/Debian.yml
+++ b/ansible-scylla-loader/tasks/Debian.yml
@@ -57,12 +57,6 @@
     force_apt_get: yes
   become: true
 
-- name: Correct java version selected
-  alternatives:
-    name: java
-    path: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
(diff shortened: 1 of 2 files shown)


vladzcloudius:
@tarzanek the OSS molecule failed due to https://github.com/scylladb/scylladb/issues/16427

PR #328, opened 2024-01-21
ansible-scylla-node: Enforces the installation of each APT key set

ebenzecri:
Before this patch, if a key with a specific ID already installed was updated (because it was expired, for example), the key won't be locally updated causing and APT failure.

This patch ensures the usage of the latest apt key available.

### DIFF
--- a/ansible-scylla-node/tasks/Debian.yml
+++ b/ansible-scylla-node/tasks/Debian.yml
@@ -8,6 +8,19 @@
       update_cache: yes
     when: install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)
 
+  - name: "Purge keyring '{{ scylla_repo_keyringfile }}'"
+    ansible.builtin.file:
+      path: "{{ scylla_repo_keyringfile }}"
+      state: absent
+    when: install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)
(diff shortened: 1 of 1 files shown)


vladzcloudius (changes requested):
Please, file a GH issue describing the issue that this PR is fixing. Please, make sure to provide steps to reproduce.

Also note that CI, has failed too.

vladzcloudius:
> Please, file a GH issue describing the issue that this PR is fixing. Please, make sure to provide steps to reproduce.
> 
> Also note that CI, has failed too.

@ebenzecri any progress with fixing the CI? You are using a `gpg` executable that is not installed in the Molecule docker image  - hence the failure.

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -1,6 +1,19 @@
 ---
 - name: Add Scylla repos
   block:
+  - name: "Purge keyring '{{ scylla_repo_keyringfile }}'"
+    ansible.builtin.file:
+      path: "{{ scylla_repo_keyringfile }}"
+      state: absent
+    when: install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)
+
+  - name: Remove an apt key by id previously added
```
vladzcloudius, line 10: You need to move both these tasks to be after the one that installs `gnupg2` (line 17 below).

vladzcloudius (changes requested):

PR #331, opened 2024-01-23
Monitoring role custom alerts

vladzcloudius:


vladzcloudius:
FYI, @ebenzecri

ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -141,7 +141,7 @@
     remote_src: true
     dest: "{{ scylla_monitoring_prom_rules_path }}/prometheus.rules.yml"
     mode: '0644'
-  when: prometheus_rules_file is defined and prometheus_rules_file == 'default' and prom_rules_file.stat.exists
+  when: prom_rules_file.stat.exists
```
igorribeiroduarte, line 144: If we always want to install the default prom_rules file, I think we can remove this condition completely, since I'd prefer this task to fail in case the file is not there.

vladzcloudius: There has been a change in the location of this file - at some point it moved to be under `prom_rules`.

You can find a similar task for a "new" location.

And we've been always installing the base file since 981bd80b5496db398e7cbcac259858f8cae46025.

This patch simply makes the logic consistent across the the whole playbook.

The sanity check you are suggesting is technically going to check if the Monitoring release is broken or not. 
One can say that it is not going to hurt. However another could argue that it's not Role's responsibility to verify that...

In any case this is outside the scope of this PR.

igorribeiroduarte: Ok

igorribeiroduarte (approved):

PR #332, opened 2024-01-25
ansible-scylla-node: Update gpg keys before running apt update

igorribeiroduarte:
The task installing gnupg2 runs the equivalent to `apt update` before moving forward
due to `update_cache` being set to `true`. However if this role is being executed for
an old cluster which has an expired key, the `apt update` will fail.
In order to prevent that, this patch does the following:
  * Removes keyring file only if gnupg2 is already installed
  * Removes scylla repo keys only if gnupg2 is already installed
  * Installs gnupg2 only if it's not installed before setting up the keys

### DIFF
--- a/ansible-scylla-node/tasks/Debian.yml
+++ b/ansible-scylla-node/tasks/Debian.yml
@@ -1,25 +1,25 @@
 ---
 - name: Add Scylla repos
   block:
-  - name: Install gnupg2 dependency
-    apt:
-      name: "gnupg2"
-      state: present
-      update_cache: yes
-    when: install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)
-
   - name: "Purge keyring '{{ scylla_repo_keyringfile }}'"
     ansible.builtin.file:
       path: "{{ scylla_repo_keyringfile }}"
       state: absent
-    when: install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)
+    when: "'gnupg2' in ansible_facts.packages and install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)"
 
   - name: Remove an apt key by id previously added
     ansible.builtin.apt_key:
       id: "{{ item }}"
       state: absent
     with_items: "{{ scylla_repo_keys }}"
-    when: install_type == 'online' and scylla_repo_keyserver is defined and scylla_repo_keys is defined and (scylla_repo_keys|length > 0)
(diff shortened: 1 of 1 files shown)


vladzcloudius:
> The task installing gnupg2 runs the equivalent to `apt update` before moving forward due to `update_cache` being set to `true`. However if this role is being executed for an old cluster which has an expired key, the `apt update` will fail. In order to prevent that, this patch does the following:
> 
>     * Removes keyring file only if gnupg2 is already installed
> 
>     * Removes scylla repo keys only if gnupg2 is already installed
> 
>     * Installs gnupg2 only after removing the keys and file mentioned above and only if it's not already installed

This makes no sense: on the one hand you say that you can only remove keys only if `gnupg2` is already installed, but then you want to install `gnupg2` only after keys were removed but only if it isn't already installed. But it can't be not installed if you have removed the keys.

Bottom line: if you were able to remove keys this means that `gnupg2` is definitely installed - hence no point in installing it in this case.

> 
>     * Adds a task to update gnupg2 at the end of the process

vladzcloudius (changes requested):
Same logical problem in the patch description.

ansible-scylla-node/tasks/Debian.yml
```diff
@@ -61,6 +61,13 @@
     with_items: "{{ scylla_deb_repos }}"
     when: install_type == 'online' and item.split(".")[-1] == "list"
 
+  - name: Update gnupg2 dependency if necessary
```
vladzcloudius, line 64: Why is this needed?

vladzcloudius (changes requested):

ansible-scylla-node/tasks/Debian.yml
igorribeiroduarte: Without this `gnupg2` will never be updated by the role, since now the previous task installing `gnupg2` will only be executed if it's not already installed.

vladzcloudius: Sure and this is how it was so far. Why is this a problem?

igorribeiroduarte: Fixed

igorribeiroduarte:
> > The task installing gnupg2 runs the equivalent to `apt update` before moving forward due to `update_cache` being set to `true`. However if this role is being executed for an old cluster which has an expired key, the `apt update` will fail. In order to prevent that, this patch does the following:
> > ```
> > * Removes keyring file only if gnupg2 is already installed
> > 
> > * Removes scylla repo keys only if gnupg2 is already installed
> > 
> > * Installs gnupg2 only after removing the keys and file mentioned above and only if it's not already installed
> > ```
> 
> This makes no sense: on the one hand you say that you can only remove keys only if `gnupg2` is already installed, but then you want to install `gnupg2` only after keys were removed but only if it isn't already installed. But it can't be not installed if you have removed the keys.
> 
> Bottom line: if you were able to remove keys this means that `gnupg2` is definitely installed - hence no point in installing it in this case.
> 
> > ```
> > * Adds a task to update gnupg2 at the end of the process
> > ```

Fixed

vladzcloudius (approved):

PR #333, opened 2024-01-25
ansible-scylla-node: Prevents APT to be updated when gnupg2 is installed

ebenzecri:
If by any chance the GPG signature is expired, it won't allow the role to install gnupg2 because "update_cache" will force APT to update its cache.


ebenzecri:
Closing this PR because https://github.com/scylladb/scylla-ansible-roles/pull/332 solves the issue.

PR #340, opened 2024-02-17
scylla-node: fix scylla_setup call

### LINKED ISSUE #339: [scylla-node]: scylla_setup is never called
HEAD: 4cdc311cdefd7f86a175642c01d0a4f367b4099d

*Description*
The following task is never called (from `common.yml`):

```
- name: configure Scylla
  shell: |
    scylla_setup --no-raid-setup --nic {{ scylla_nic }} --setup-nic-and-disks --no-ec2-check --no-kernel-check --no-ntp-setup --no-io-setup --no-node-exporter --no-coredump-setup --no-sysconfig-setup --no-swap-setup 
  become: true
  when: ansible_facts.services["scylla-server.service"] is defined and ansible_facts.services["scylla-server.service"]["state"] != "running"
```

Reason:
* When a new nodes is being deployed `ansible_facts.services["scylla-server.service"] is defined` is false because we populate services facts early in the play.
* When Role is applied to an existing node then `ansible_facts.services["scylla-server.service"]["state"] != "running"` is false.

vladzcloudius:
Fix an issue with scylla_setup invocation.
Remove an old perftune.py bug workaround.

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -147,9 +147,8 @@
 
 - name: configure Scylla
   shell: |
-    scylla_setup --no-raid-setup --nic {{ scylla_nic }} --setup-nic-and-disks --no-ec2-check --no-kernel-check --no-ntp-setup --no-io-setup --no-node-exporter --no-coredump-setup --no-sysconfig-setup --no-swap-setup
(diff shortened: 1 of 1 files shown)


vladzcloudius:
Don't merge - I'm adding another patch.

PR #342, opened 2024-02-19
scylla-node: use rpc_address instead of listen_address on wait CQL port to be up

aurelilys:
Hello,

I recently wanted to deploy a scylladb cluster with this playbook, I encountered a small problem when I redefined the scylla_rpc_address varriable to specific IP, the start_one_node task waits for scylla to be up on the wrong address (listen_address) and never succeeds.

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -314,6 +314,11 @@
     listen_address: "{{ scylla_listen_address }}"
 
 # The same relates to the below
+- name: Resolve scylla_rpc_address
+  set_fact:
+    rpc_address: "{{ scylla_rpc_address }}"
+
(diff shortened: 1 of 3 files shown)


vladzcloudius (changes requested):
The current behavior of the Role is a correct one since we are waiting for the CQL port on the node itself (note the `delegate_to` statement).

I suggest you open GH issue and describe the issue that you had with all relevant details.

vladzcloudius:
> The current behavior of the Role is a correct one since we are waiting for the CQL port on the node itself (note the `delegate_to` statement).
> 
> I suggest you open GH issue and describe the issue that you had with all relevant details.

I'm taking it back - my apologies. 
Your patch is good!

vladzcloudius (approved):

PR #345, opened 2024-02-19
wipe_node.yml: Add a playbook for completely cleaning a former scylla node

igorribeiroduarte:


vladzcloudius (changes requested):
Let's remove all scylla packages as a last step.
This is should be trivial using https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html

igorribeiroduarte:
> Let's remove all scylla packages as a last step. This is should be trivial using https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_module.html

Done, @vladzcloudius

PR #347, opened 2024-02-28
all-roles: use HTTPS for accessing downloads.scylladb.com

vladzcloudius:
Starting March 5 http access to downloads.scylladb.com is going to be blocked. Let's switch to using https.

Ref scylladb/scylla-pkg#3764


tarzanek:
lgtm, merging

PR #354, opened 2024-03-12
Monitoring dashboards auto detection

vladzcloudius:
Auto-detect dashboards versions unless explicitly specified.

### DIFF
--- a/ansible-scylla-monitoring/tasks/common.yml
+++ b/ansible-scylla-monitoring/tasks/common.yml
@@ -217,3 +217,26 @@
     dest: "{{ scylla_monitoring_config_path }}/rule_config.yml"
     mode: '0644'
   when: rule_config_stat.stat.exists|bool == false
+
+- name: Set _scylla_monitoring_params to scylla_monitoring_params
+  set_fact:
+    _scylla_monitoring_params: "{{ scylla_monitoring_params | default('') }}"
+
+- name: Auto detect Scylla Manager version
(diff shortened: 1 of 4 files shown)


vladzcloudius:
Let me fix the CI

ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -217,3 +217,26 @@
...
+    - name: Adding Scylla Manager dashboard option if getting a Manager version was successful
+      set_fact:
+        _scylla_monitoring_params: "{{ _scylla_monitoring_params }} -M {{ _sm_version_out.stdout }}"
+      when: _sm_version_out.rc == 0
+
+    - name: Error out if failed to get Scylla Manager version
+      fail:
+        msg: "Can't auto-detect Scylla Manager dashboard version. Either specify it explicitly or install Scylla Manager Server on the Scylla Manager host."
+      when: _sm_version_out.rc != 0
+  when: _scylla_monitoring_params is not search("-M\s+")
```
igorribeiroduarte, line 242: Why don't you just check if it's empty?
I assume you're just trying to validate if the user passed a valid value, but this shouldn't be responsibility of this task.

vladzcloudius: `_scylla_monitoring_params` is not going to be empty. It can have other parameters like a username and/or password

PR #364, opened 2024-03-15
wipe_node.yml: Remove configuration files

igorribeiroduarte:


example-playbooks/wipe_node/wipe_node.yml
```diff
@@ -92,7 +92,5 @@
       shell: |
         systemctl daemon-reload
 
-    - name: Remove Scylla packages
```
vladzcloudius, line 95: Please, add more "why?" text to a patch description. 
AFAIU this patch is needed because there is some issue with the `package` Ansible module, isn't it?

vladzcloudius (changes requested):

example-playbooks/wipe_node/wipe_node.yml
igorribeiroduarte: > Please, add more "why?" text to a patch description.

Done.

>  AFAIU this patch is needed because there is some issue with the package Ansible module, isn't it?

Yes, the `package` module doesn't allow us to remove configuration files. I've added a more detailed explanation in the patch description.

vladzcloudius (approved):

PR #375, opened 2024-05-10
scylla-node/Debian_install.yml: Don't check if the installed scylla v…

vladzcloudius:
…ersion appears in the repo exactly once

We have a check that validates that a value in `scylla_version` matches exactly one (full) scylla verison from the repository so that the Role can install it.

However this check is only relevant when we are going to install anything, and when scylla is already installed - we are not going to. Hence, this check is only making noise in such a case.

One common example of such a situation is when a custom build is installed.

### DIFF
--- a/ansible-scylla-node/tasks/Debian_install.yml
+++ b/ansible-scylla-node/tasks/Debian_install.yml
@@ -19,7 +19,7 @@
     - name: "Validate scylla version correctness"
       ansible.builtin.fail:
         msg: "Too many/few choices for a requested version '{{ scylla_version_to_install }}': {{ aptversions.stdout_lines }}. Bailing out!"
-      when: aptversions.stdout_lines | length != 1
+      when: (not scylla_is_installed or upgrade_version) and aptversions.stdout_lines | length != 1
 
     - name: If Scylla is installed check if it's the same version as requested
       fail:


ebenzecri (approved):

ansible-scylla-node/tasks/Debian_install.yml
```diff
@@ -19,7 +19,7 @@
     - name: "Validate scylla version correctness"
       ansible.builtin.fail:
         msg: "Too many/few choices for a requested version '{{ scylla_version_to_install }}': {{ aptversions.stdout_lines }}. Bailing out!"
-      when: aptversions.stdout_lines | length != 1
+      when: not scylla_is_installed and aptversions.stdout_lines | length != 1
```
igorribeiroduarte, line 22: You need to add `and not upgrade_version` to your condition, because we still want to run this task to be executed if we're performing an upgrade

igorribeiroduarte (changes requested):

ansible-scylla-node/tasks/Debian_install.yml
vladzcloudius: Fixed, please, re-check, @igorribeiroduarte

igorribeiroduarte (approved):

PR #380, opened 2024-05-16
replace_node.yml: Allow disabling RBNO for the new node during replace

igorribeiroduarte:
When we're working in a cluster with mixed CPUSETs, it might be necessary to disable RBNO during a replace to avoid repair reader timeouts.
Having that in mind, this PR adds an option to disable RBNO only during the replace.


vladzcloudius (approved):
Looks good. Just add the PR motivation in its description and we can merge.

PR #381, opened 2024-05-16
replace_node.yml: Fix small bugs in the playbook

igorribeiroduarte:


vladzcloudius (changes requested):
Neither a description nor the code makes sense to me.
Why this patch is needed?

vladzcloudius (approved):

PR #388, opened 2024-07-01
example_playbooks/run_cleanup: properly remove just hash of directory

### LINKED ISSUE #387: cleanup script will fail if table has "-" in its name
https://github.com/scylladb/scylla-ansible-roles/blob/master/example-playbooks/run_cleanup/files/cleanup.sh

```
...
error: Scylla API server HTTP POST to URL '/storage_service/keyspace_cleanup/alternator_my-keyspace_dev' failed: Can't find a column family my in keyspace alternator_my-keyspace_dev
...
```
name of table is "my-table"

tarzanek:
instead of only getting first part until "-"

cql cannot have in table name "-" but
alternator can, so we cannot assume
that just first part until "-" is table name
logic has to be the other way, only whatever is after last "-" has to be deleted, all before is table name

fixes #387

### DIFF
--- a/example-playbooks/run_cleanup/files/cleanup.sh
+++ b/example-playbooks/run_cleanup/files/cleanup.sh
@@ -27,7 +27,8 @@ echo "cleaning first detected data dir:"; pwd
 PATHS=$(du -m --max-depth=2 .|sort -k1h|awk -F"/" 'NF==3'|awk '{ print $NF}'|grep -E -v '^./system/'\|'^./system_auth/'\|'^./system_distributed/'\|'^./system_traces/'\|'^./system_schema/')
 
 for i in $PATHS; do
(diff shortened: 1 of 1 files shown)


tarzanek:
tested with an alternator table as per issue

script log:
```
ubuntu@ubuntu:~$ ./cleanup.sh
running on ubuntu
cleaning first detected data dir:
/var/lib/scylla/data
evaluating /var/lib/scylla/data/./alternator_my-usertable/my-usertable-b2a9512037af11efbdddb8c11b2d3d7b/
cleanup of alternator_my-usertable my-usertable
...
```

scylla log:
```
...
Jul 01 13:45:19 ubuntu scylla[849]:  [shard 0:stre] api - force_keyspace_cleanup: keyspace=alternator_my-usertable tables={table{name=my-usertable, id=b2a95120-37af-11ef-bddd-b8c11b2d3d7b}}
...
```

igorribeiroduarte (approved):

PR #390, opened 2024-07-03
replace_node.yml: Add any_errors_fatal to the replace_node playbook

igorribeiroduarte:


tarzanek:
ok, safety and stopping whenever the role has a problem, ok, let's merge

PR #394, opened 2024-07-05
ansible-scylla-node: Allow auto-detecting nvme cards

igorribeiroduarte:
This patch adds a `detect_nvmes` field, which when set to True will use 'nvme list' command for getting the list of nvme devices in the machine and pass all of them to scylla_raid_setup.

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -190,6 +190,7 @@ scylla_dependencies:
   - curl
   - wget
   - python3-yaml # this will fail on non centos systems
+  - nvme-cli
   #- software-properties-common
   #- apt-transport-https
   #- gnupg2
@@ -423,6 +424,10 @@ scylla_manager_agent_config: |
   # - /dev/nvme0n2
   # - /dev/nvme0n3
 scylla_raid_online_discard: False
+
+# If scylla_raid_setup is not given and this is set to True, the role will use 'nvme list' for getting the list of nvme devices
+# in the machine and pass all of them to scylla_raid_setup
+detect_nvmes: False
 
 skip_coredump: False
 devmode: False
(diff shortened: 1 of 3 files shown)


ansible-scylla-node/tasks/common.yml
```diff
@@ -48,6 +48,18 @@
       when: cpuset_command is defined and skip_cpuset is defined and skip_cpuset|bool == false
   become: true
 
+- name: Detect NVME disks
+  block:
+    - name: Get list of NVMEs
+      shell: |
+        nvme list | awk '{print $1}' | tail -n +3
```
vladzcloudius, line 55: `nvme` is not installed by default. You need to install it before you can use it.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/common.yml
vladzcloudius: On Ubuntu this is an `nvme-cli` package for example.

vladzcloudius: Other than that - very nice idea.

igorribeiroduarte: I'm now installing `nvme-cli` for both, Debian and RH.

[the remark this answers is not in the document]
igorribeiroduarte: `lsblk` would show non-nvme disks as well.
I've added `nvme-cli` to the list of dependencies in the role.

ansible-scylla-node/defaults/main.yml
```diff
@@ -190,6 +190,7 @@ scylla_dependencies:
   - curl
   - wget
   - python3-yaml # this will fail on non centos systems
```
vladzcloudius, line 192: Unrelated to this PR: this comment is BS ;)

ansible-scylla-node/tasks/RedHat.yml
```diff
@@ -120,6 +120,11 @@
     for i in `yum search python3|grep -i pyyaml|awk '{ print $1 }'`; do sudo yum -y install $i; done
   become: true
 
+- name: install nvme-cli
+  shell: |
+    yum install -y nvme-cli
```
vladzcloudius, line 125: I wonder why we don't use `scylla_dependencies` for RH too?

vladzcloudius (approved):

PR #397, opened 2024-07-16
replace_node.yml: Cleanup job aliases for 'nodetool drain' and 'stop scylla service' async tasks

igorribeiroduarte:
As explained in the documentation of the async module { https://github.com/scylladb/scylla-ansible-roles/tree/master/example-playbooks/async_extra#failures }, if an async job finishes with a bad result the task will not cleanup the job alias. So we need a separate task for doing that.

### DIFF
--- a/example-playbooks/replace_node/replace_node.yml
+++ b/example-playbooks/replace_node/replace_node.yml
@@ -30,26 +30,40 @@
           become: true
 
         - name: Run nodetool drain
-          async_task:
-            shell: |
-              nodetool drain
-            alias: drain_scylla
-            async: 60
-            retries: 60
-            delay: 1
-          ignore_errors: true
-          register: _drain_scylla_output
+          block:
+            - name: Async nodetool drain
+              async_task:
+                shell: |
+                  nodetool drain
+                alias: drain_scylla
+                async: 60
+                retries: 60
+                delay: 1
+              ignore_errors: true
+              register: _drain_scylla_output
+
+            - name: Cleanup job alias
+              shell: |
+                rm -rf "{{ _drain_scylla_output.alias_path }}"
+              when: _drain_scylla_output.failed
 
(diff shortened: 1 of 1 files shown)


example-playbooks/replace_node/replace_node.yml
```diff
@@ -30,26 +30,38 @@
...
+                alias: drain_scylla
+                async: 60
+                retries: 60
+                delay: 1
+              ignore_errors: true
+              register: _drain_scylla_output
+
+            - name: Cleanup job alias
+              shell: |
+                rm -rf "{{ _drain_scylla_output.alias_path }}"
```
vladzcloudius, line 47: Don't you want to do that only if the task fails? I'm sure there is a field in `_drain_scylla_output` that has the exit status, no?

example-playbooks/replace_node/replace_node.yml
```diff
@@ -30,26 +30,38 @@
...
+                alias: stop_scylla
+                async: 5
+                retries: 5
+                delay: 1
+              ignore_errors: true
+              register: _stop_scylla_output
+
+            - name: Cleanup job alias
+              shell: |
+                rm -rf "{{ _stop_scylla_output.alias_path }}"
```
vladzcloudius, line 64: Same question here.

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: Fixed

vladzcloudius (approved):

PR #398, opened 2024-07-26
replace_node.yml: Allow ignoring dead nodes during replace

igorribeiroduarte:
Scylla has an `ignore_dead_nodes_for_replace` config which allows the user to ignore dead nodes during the replace and so far it was not supported by this playbook.
This PR adds support to this feature by adding a variable `dead_nodes_to_ignore`.

### DIFF
--- a/example-playbooks/replace_node/replace_node.yml
+++ b/example-playbooks/replace_node/replace_node.yml
@@ -312,6 +312,7 @@
         state: restarted
         enabled: yes
       become: true
+      when: inventory_hostname == new_node
 
     - name: Re-enable RBNO for the new node
       block:
@@ -326,7 +327,7 @@
           shell: |
             pkill -1 scylla$
       become: true
-      when: disable_rbno|bool and _rbno_enabled is defined and _rbno_enabled|bool
+      when: inventory_hostname == new_node and disable_rbno|bool and _rbno_enabled is defined and _rbno_enabled|bool
 
 
 - name: Resume scylla-manager tasks
--- a/example-playbooks/replace_node/replace_node.yml
+++ b/example-playbooks/replace_node/replace_node.yml
@@ -173,6 +173,32 @@
       delegate_to: "{{ temporary_seed }}"
       run_once: true
 
+    - name: Set ignore_dead_nodes_for_replace using host ids
+      block:
+        - name: Set dead nodes host ids
+          set_fact:
+            _dead_nodes_host_ids: "{{ _dead_nodes_host_ids | default([]) + [item.value] }}"
+          when: item.key in dead_nodes_to_ignore
+          loop: "{{ _host_ids.json }}"
+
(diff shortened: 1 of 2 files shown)


tarzanek (approved):

example-playbooks/replace_node/replace_node.yml
```diff
@@ -164,16 +181,6 @@
...
-        - name: Get the host id for all nodes
-          uri:
-            url: "http://{{ scylla_api_address }}:{{ scylla_api_port }}/storage_service/host_id"
-            method: GET
-          register: _host_ids
-          until: _host_ids.status == 200
-          retries: 5
-          delay: 1
-          delegate_to: "{{ temporary_seed }}"
-
```
vladzcloudius, line 176: Removing this block will break the `set_fact` below because you are iterating over a variable from a different play (`_host_ids`).

And according to the documentation "registered" variables scope is a corresponding play: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#registering-variables

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
```diff
@@ -256,6 +271,13 @@
         state: absent
       when: _replace_node_first_boot_grep.failed
 
+    - name: Remove the ignore_dead_nodes_for_replace record
+      lineinfile:
+        path: /etc/scylla/scylla.yaml
+        regexp: '^ignore_dead_nodes_for_replace:'
+        state: absent
+      when: dead_nodes_to_ignore|length > 0
```
vladzcloudius, line 279: What if we are using a `scylla` version that expects IPs/hostnames in `ignore_dead_nodes_for_replace`? For example, 2022.2.19?

Those would be same versions that use `replace_address_first_boot` and not the new `replace_node_first_boot`

vladzcloudius (changes requested):

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: The documentation actually says that they are valid for the rest of the playbook, including subsequent plays within the same playbook run:
> Registered variables are stored in memory. You cannot cache registered variables for use in future playbook runs. Registered variables are only valid on the host for the rest of the current playbook run, including subsequent plays within the same playbook run.

igorribeiroduarte: Anyway, this task is not being moved to the other play anymore

example-playbooks/replace_node/replace_node.yml
igorribeiroduarte: I've adjusted the PR to also support scylla versions that still use addresses instead of host ids in `ignore_dead_nodes_for_replace`

vladzcloudius (approved):

PR #402, opened 2024-08-15
scylla-ansible-roles: Adds example playbook "kernel_version_enforcer"

ebenzecri:
"kernel_version_enforcer" playbook allow the user to:
- Pin a specific kernel version (and ensure it will be picked in the next reboot) if required
- Upgrade kernel version to the latest available
- Purge all old kernel versions
- Upgrade all upgradable packages

### DIFF
--- a/example-playbooks/kernel_version_enforcer/ubuntu/grub.yml
+++ b/example-playbooks/kernel_version_enforcer/ubuntu/grub.yml
@@ -1,0 +1,30 @@
+---
+
+- name: Get {{ grub_config_file }} metadata
+  ansible.builtin.stat:
+    path: "{{ grub_config_file }}"
+  register: grub_config
+
+- name: Fail if GRUB config file doesn't exist
+  ansible.builtin.fail:
+    msg: "{{ grub_config_file }} doesn't exist"
+  when: not grub_config.stat.exists
+
+- name: Get GRUB entries
+  ansible.builtin.command: grep -E "^\smenuentry" {{ grub_config_file }}
+  register: grub_entries
+
+- name: Get GRUB index for '{{ image_package_prefix }}-{{ image_version }}'
+  ansible.builtin.set_fact:
+    target_grub_index="{{ grub_index }}"
+  when:
+    - image_version in item
+    - not "recovery mode" in item
+    - target_grub_index is not defined
+  loop: "{{ grub_entries.stdout_lines }}"
+  loop_control:
+    index_var: grub_index
+
+- name: Set index '1>{{ target_grub_index }}' to be used in the next reboot
+  ansible.builtin.command: grub-reboot "1>{{ target_grub_index }}"
+  become: true
--- a/example-playbooks/kernel_version_enforcer/ubuntu/main.yml
+++ b/example-playbooks/kernel_version_enforcer/ubuntu/main.yml
@@ -1,0 +1,134 @@
+---
+
+- name: Get current kernel image version
+  ansible.builtin.command: uname --kernel-release
+  register: uname_pre_output
+
+- name: Save kernel image version
+  ansible.builtin.set_fact:
+    detected_image_version="{{ uname_pre_output.stdout_lines | first }}"
+
+- name: Define if the kernel image should be installed
+  ansible.builtin.set_fact:
+    kernel_image_required="{{ image_version is version(detected_image_version, 'ne') or upgrade_latest_kernel }}"
+
+- name: Ensure apt is ready to be used
+  ansible.builtin.apt:
+    update_cache: true
+    autoclean: true
+    autoremove: true
+    force_apt_get: true
+  become: true
+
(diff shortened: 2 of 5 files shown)


vladzcloudius (changes requested):
Eduardo, as we discussed, it would be better to have this functionality integrated into the Node Role.

example-playbooks/kernel_version_enforcer/ubuntu/main.yml
```diff
@@ -0,0 +1,118 @@
...
+  ansible.builtin.include_tasks: grub.yml
+  when:
+    - reconfiguration_required
+    - not upgrade_latest_kernel
+
+- name: Stop, reboot and start each node (if required)
+  ansible.builtin.include_tasks: stop_reboot_start.yml
+  when: reconfiguration_required
+
+- name: Set final kernel image version if '{{ image_version }}' was installed
```
vladzcloudius, line 89: I don't understand why all tasks below this one are required to begin with.

example-playbooks/kernel_version_enforcer/ubuntu/main.yml
```diff
@@ -0,0 +1,118 @@
...
+- name: Set final kernel image version if the latest one was installed
+  ansible.builtin.set_fact:
+    final_image_version="{{ target_image_version }}"
+  when:
+    - reconfiguration_required
+    - upgrade_latest_kernel
+
+- name: Enforce kernel version '{{ final_image_version }}' usage
+  ansible.builtin.include_tasks: kernel_enforce_cleanup.yml
+  when: reconfiguration_required
```
vladzcloudius, line 103: Why not using `apt autoremove`?

vladzcloudius (changes requested):

example-playbooks/kernel_version_enforcer/ubuntu/stop_reboot_start.yml
```diff
@@ -0,0 +1,122 @@
...
+- name: Check if Scylla is installed
+  ansible.builtin.set_fact:
+    scylla_installation="{{ true if ansible_facts.services['scylla-server.service'] is defined else false }}"
+
+- name: Stop Scylla
+  block:
+    - name: Mask scylla-server service
+      ansible.builtin.systemd:
+        name: scylla-server
+        masked: true
```
vladzcloudius, line 14: "masked"? Why?

example-playbooks/kernel_version_enforcer/ubuntu/stop_reboot_start.yml
```diff
@@ -0,0 +1,122 @@
+---
+- name: Populate service facts
+  ansible.builtin.service_facts:
+
+- name: Check if Scylla is installed
+  ansible.builtin.set_fact:
+    scylla_installation="{{ true if ansible_facts.services['scylla-server.service'] is defined else false }}"
+
+- name: Stop Scylla
```
vladzcloudius, line 9: We should stop duplicating "Scylla Rolling Restart" versions and have one that is used everywhere.
The best version of Scylla shutdown can be found in https://github.com/scylladb/scylla-ansible-roles/blob/master/example-playbooks/rolling_ops/rolling_custom_build_install.yml.

We should add the missing "rescue" block to https://github.com/scylladb/scylla-ansible-roles/blob/master/example-playbooks/rolling_ops/rolling_restart.yml and then it would be good to be used everywhere.

example-playbooks/kernel_version_enforcer/ubuntu/grub.yml
```diff
@@ -0,0 +1,30 @@
+---
```
vladzcloudius, line 1: Procedure below doesn't seem to be standard.
Here is an example of a procedure I'm aware of: https://askubuntu.com/questions/100232/how-do-i-change-the-grub-boot-order.

In gist you need to change `GRUB_DEFAULT` value in `/etc/default/grub` and then run `sudo update-grub`.

vladzcloudius (changes requested):

example-playbooks/kernel_version_enforcer/ubuntu/stop_reboot_start.yml
vreniers: This stop and start functionality, is indeed preferably implemented only once (DRY principle). 
Perhaps as a task in `ansible-scylla-node` role, and perhaps it even makes sense to group such tasks in a folder. There is already such a task present `start_one_node.yml`.

example-playbooks/kernel_version_enforcer/ubuntu/main.yml
```diff
@@ -0,0 +1,118 @@
+---
```
vreniers, line 1: I don't think this functionality should be inside a playbook. We will want to re-use this functionality when setting up a ScyllaDB node.

vreniers (changes requested):

PR #405, opened 2024-09-17
ansible-scylla-monitoring: Always look for the prometheus_rules_file in scylla_monitoring_role_config_path

igorribeiroduarte:


ansible-scylla-monitoring/defaults/main.yml
```diff
@@ -20,6 +20,7 @@ scylla_manager_ip: '127.0.0.1'
 # Download the scylla-monitoring from here
 scylla_monitoring_archive_url: 'https://github.com/scylladb/scylla-monitoring/archive/scylla-monitoring-3.10.1.tar.gz'
 
+scylla_monitoring_local_files_path: "{{ inventory_dir }}"
```
vladzcloudius, line 23: What is "local files"? 
Why not calling it `scylla_monitoring_configuration_files_path` ?

vladzcloudius (changes requested):

ansible-scylla-monitoring/defaults/main.yml
igorribeiroduarte: My idea was to clarify that this path is where the role will look for local files only and it's not related to the paths where these files will be copied to, in the remote nodes

igorribeiroduarte: btw, we already have a `scylla_monitoring_config_path` variable, so adding a `scylla_monitoring_configuration_files_path` (which is exactly the remote path that the config files will be copied to) could cause some confusion

vladzcloudius (changes requested):

ansible-scylla-monitoring/defaults/main.yml
igorribeiroduarte: Done

ansible-scylla-monitoring/defaults/main.yml
```diff
@@ -20,6 +20,7 @@ scylla_manager_ip: '127.0.0.1'
 # Download the scylla-monitoring from here
 scylla_monitoring_archive_url: 'https://github.com/scylladb/scylla-monitoring/archive/scylla-monitoring-3.10.1.tar.gz'
 
+scylla_monitoring_role_config_path: "{{ inventory_dir }}"
```
vladzcloudius, line 23: Please, put a description for a new parameter

vladzcloudius (changes requested):

ansible-scylla-monitoring/defaults/main.yml
igorribeiroduarte: Done

vladzcloudius (approved):

PR #411, opened 2024-10-11
Compatibility with newer Ansible versions (tested 2.17.5)

### LINKED ISSUE #409: [monitoring] Ubuntu 24 as image for monitoring VM errors with docker install
```
TASK [ansible-scylla-monitoring : add docker's apt key] *****************************************************************************************************************************************
task path: /home/scyllaadm/azure-test/scylla-ansible-roles/ansible-scylla-monitoring/tasks/Debian.yml:36
fatal: [172.171.241.170]: FAILED! => {"changed": false, "msg": "Failed to download key at https://download.docker.com/linux/ubuntu/gpg: An unknown error occurred: 'CustomHTTPSConnection' object has no attribute 'cert_file'"}
```

so it seems docker install is likely different on U24
on U22 above passes OK

vreniers:
This patch introduces several fixes to allow compatibility with newer ansible versions.
Ansible versions 2.14 or older are EOL ([source](https://docs.ansible.com/ansible/devel/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix))

The patch introduces the following:
- Patches the async plugin library
- Patches scylla-ansible-manager to properly pass variables to the scylla-ansible-node role with newer Ansible versions.
- Patches a bug in deactivation of services: Services were listed as a fact but with a 'non-found' status.
- Detects errors when scylla-manager does not start.

Fixes #409: A newer Ansible version was required to execute properly with Ubuntu 24 which came with a newer Python version out-of-the-box, which was not supported by older Ansible versions.

### DIFF
--- a/ansible-scylla-manager/tasks/main.yml
+++ b/ansible-scylla-manager/tasks/main.yml
@@ -1,24 +1,4 @@
 ---
-- name: deploy local Scylla on the Manager node
-  import_role:
-    name: "{{ role_path }}/../ansible-scylla-node"
-  vars:
-  # TODO how to use scylla_manager_db_vars to be passed on as role params?
-    install_only: True
-    scylla_manager_enabled: false
-    scylla_version: 'latest'
-    scylla_edition: "{{ scylla_manager_db_vars.scylla_edition|default('oss') }}"
-    elrepo_kernel: false
-    scylla_repo_keyserver: "{{ scylla_manager_db_vars.scylla_repo_keyserver|default('') }}"
-    scylla_repo_keys: "{{ scylla_manager_db_vars.scylla_repo_keys|default([]) }}"
-    scylla_dependencies: "{{ scylla_manager_db_vars.scylla_dependencies|default([]) }}"
-    scylla_ssl:
-      internode:
-        enabled: false
-      client:
-        enabled: false
-
-
 - name: install Scylla Manager
   include_tasks: "{{ ansible_os_family }}.yml"
 
--- a/ansible-scylla-node/tasks/manager_agents.yml
+++ b/ansible-scylla-node/tasks/manager_agents.yml
@@ -61,7 +61,3 @@
     enabled: yes
   become: true
   when: manager_agent_config_change.changed and start_scylla_service is defined and start_scylla_service|bool
-  ignore_errors: true
-  #TODO: remove ignore_errors when ansible is bumped to 2.10.4 or 2.9.16 as per https://github.com/ansible/ansible/issues/71528
-
-
--- a/example-playbooks/async_extra/action_plugins/async_alias.py
+++ b/example-playbooks/async_extra/action_plugins/async_alias.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
 # -*- coding: utf-8 -*-
 #
 # Copyright (C) 2021 ScyllaDB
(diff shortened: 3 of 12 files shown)


example-playbooks/async_extra/action_plugins/async_alias.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
```
vladzcloudius, line 1: The previous version is actually the one recommended. 
Why did you have to change it?

vladzcloudius (changes requested):

vladzcloudius:
@tarzanek PR looks good to me except for the shabang question above.
Let's estimate the blast radius after we merge it: I'd like to upgrade all our toolchain to latest versions ASAP.
For example, in the cloud-access-wrapper PR you merged yesterday I had to virtually implement enum.StrEnum because we still use Python3.8.

example-playbooks/async_extra/action_plugins/async_alias.py
vladzcloudius: The doc page is referenced in the commit message

vladzcloudius (dismissed):

ansible-scylla-node/tasks/manager_agents.yml
```diff
@@ -60,8 +60,4 @@
     state: restarted
     enabled: yes
   become: true
-  when: manager_agent_config_change.changed and start_scylla_service is defined and start_scylla_service|bool
-  ignore_errors: true
-  #TODO: remove ignore_errors when ansible is bumped to 2.10.4 or 2.9.16 as per https://github.com/ansible/ansible/issues/71528
-
```
vladzcloudius, line 66: This is one of the changes that may break things with older Ansible

vladzcloudius: This is the only such change AFAICT

igorribeiroduarte (changes requested):
Just these 2 small nitpicks
Other than that, LGTM

ansible-scylla-manager/tasks/main.yml
```diff
@@ -1,24 +1,4 @@
...
-- name: deploy local Scylla on the Manager node
-  import_role:
-    name: "{{ role_path }}/../ansible-scylla-node"
-  vars:
-  # TODO how to use scylla_manager_db_vars to be passed on as role params?
-    install_only: True
-    scylla_manager_enabled: false
-    scylla_version: 'latest'
-    scylla_edition: "{{ scylla_manager_db_vars.scylla_edition|default('oss') }}"
-    elrepo_kernel: false
```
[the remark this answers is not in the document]
vreniers: I get the impression that this manager setup wants the most basic local ScyllaDB setup as possible. Ideally, there shouldn't be much to pass in terms of variables in such a case.

This variable it already happens to be false by default in the scylla-node role. Since it's not doing anything, I removed it.

I.e. I think we should align the scylla-node role to have the most 'basic' defaults as possible. Minimal hassle to get an out-of-the-box working experience.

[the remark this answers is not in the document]
vreniers: I've placed the variable back, and we can address such variables and defaults in a separate PR.

igorribeiroduarte (approved):

igorribeiroduarte (approved):

PR #412, opened 2024-10-11
ansible-scylla-node: upgrade/pre_upgrade.yml: Remove redundant code

### LINKED ISSUE #134: [Scylla Node Role]: upgrade does a lot of not needed steps that slow the whole process down
HEAD: 365a23dabb8fcd07a44654660847d37824e33a51

When `upgrade_version: true` Role is going to first shut down the CQL protocol, then a Gossip protocol, only then drain the node and then shut the node down.

While according to a procedure described here: https://docs.scylladb.com/upgrade/upgrade-enterprise/ all what's required is drain + stop.

vreniers:
This removes redundant code which is not necessary, it is sufficient to simply drain a node.

Fixes #134


tarzanek (approved):
lgtm, thank you!

PR #413, opened 2024-10-11
ansible-scylla-node: auto-detect NVMe: only use non-mounted disks.

### LINKED ISSUE #407: `detect_nvmes` feature doesn't work for AWS
AWS instances have an extra NVME for the boot disk, so when you enable `detect_nvmes` on a cluster with AWS instances, the node role will try to also use this extra NVME, since by default we try to use all the NVMEs present in the instance.
We should change the role to only use NVMEs that are not already being used.

CC: @tarzanek

vreniers:
NVMe disks should only be selected that are not in use, which should be the case when they are not mounted.

Fixes #407

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -50,9 +50,12 @@
 
 - name: Detect NVME disks
   block:
-    - name: Get list of NVMEs
+    - name: Get list of NVMEs with no mount points
       shell: |
-        nvme list | awk '{print $1}' | tail -n +3
+        for nvme_disk in $(nvme list | awk '{print $1}' | tail -n +3)
+        do
+          { echo -n "$nvme_disk "; lsblk -n "$nvme_disk" -o MOUNTPOINT | grep / | wc -l; } | awk '{if ($2=="0") print $1}'
+        done
       register: _detected_nvmes
 
     - set_fact:
@@ -79,6 +82,10 @@
         set_fact:
           _disable_online_discard: "--online-discard {{ scylla_raid_online_discard }}"
         when: _scylla_raid_setup_help.stdout is search("--online-discard")
+
+      - name: "ansible-scylla-node: common.yml: Setting raid setup with disks"
+        debug:
+          msg: "Using the following disk devices: {{ scylla_raid_setup | join(',') }}"
 
       - name: run raid setup if there is no scylla mount yet
         shell: "scylla_raid_setup --disks {{ scylla_raid_setup | join(',') }} --update-fstab {{ _disable_online_discard }}"


igorribeiroduarte (dismissed):

ansible-scylla-node/tasks/common.yml
```diff
@@ -50,9 +50,12 @@
 
 - name: Detect NVME disks
   block:
-    - name: Get list of NVMEs
+    - name: Get list of NVMEs with no mount points
       shell: |
-        nvme list | awk '{print $1}' | tail -n +3
+        for nvme_disk in $(lsblk -oNAME -nd | grep nvme)
+        do
+          { echo -n "/dev/$nvme_disk "; lsblk "/dev/$nvme_disk" -o MOUNTPOINT | grep / | wc -l; } | awk '{if ($2=="0") print $1}'
```
tarzanek, line 57: lsblk -n ?
to void printing header ?

ansible-scylla-node/tasks/common.yml
```diff
@@ -50,9 +50,12 @@
 
 - name: Detect NVME disks
   block:
-    - name: Get list of NVMEs
+    - name: Get list of NVMEs with no mount points
       shell: |
-        nvme list | awk '{print $1}' | tail -n +3
+        for nvme_disk in $(lsblk -oNAME -nd | grep nvme)
```
vladzcloudius, line 55: Is there a strong guarantee that NVME disks are always going to have an 'nvme' in their name in the output of `lsblk`?

vladzcloudius:
> ```
>   lsblk -oNAME,PATH,PTTYPE,MOUNTPOINT -nd | grep nvme | awk '{if ($3=="" && $4=="") print $2}'
> ```
> 
> misbehaves on md0 with multiple nvmes

What do you mean exactly, @tarzanek ?
Please, provide the corresponding output so that we can all see what's the problem.

vladzcloudius (changes requested):

ansible-scylla-node/tasks/common.yml
vreniers: It is the case for AWS, GCP and Azure. Regardless, that should be part of parameter documentation as this is the expectation.

vladzcloudius: The question wasn't about infra - it's an OS question. Linux to be specific. And I don't think there is such a guarantee.
I don't see any reason not to continue using the `nvme` utility which does provide you all required guarantees.

vreniers: Adjusted.

vladzcloudius (approved):

PR #425, opened 2024-11-25
scylla-node: Append 'dc_suffix' to dc name when GPFS is used

igorribeiroduarte:


ansible-scylla-node/templates/cassandra-rackdc.properties.j2
```diff
@@ -14,7 +14,7 @@
 #
 
 {% if scylla_snitch == 'GossipingPropertyFileSnitch' %}
-dc={{ dc }}
+dc={{ dc + dc_suffix | default('') }}
```
vladzcloudius, line 17: This is a somewhat unexpected syntax. I was expecting something like this:

```
dc={{ dc }}{{ dc_suffix | default('') }}
```

Did you test your version?

Issue #151, opened 2022-08-08
generate proper hostname for a SSL certificate


tarzanek:

vladzcloudius:
@tarzanek It's not clear what's the context of this issue:
* What didn't work?
* How to reproduce?

tarzanek:
the goal is to be able from role to use certificates with validate=true even without [certfiles] section exactly as in
https://docs.scylladb.com/stable/operating-scylla/security/gen-cqlsh-file.html

this can be achieved by generating self signed certs in a better way and with needed metadata

PR #427, opened 2024-12-06
ansible-scylla-node: Allow the user to use a same certificate for all the nodes

igorribeiroduarte:


tarzanek:
so with wildcard certificate support I guess https://github.com/scylladb/scylla-ansible-roles/issues/151 could be closed too

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -1,18 +1,20 @@
 ---
+
 - set_fact:
-    _localhost_cert_path: "{{ scylla_ssl.localhost_cert_path | default(inventory_dir) }}"
+    _localhost_cert_path: "{{ scylla_ssl.localhost_cert_path | default(inventory_dir + '/ssl/' + inventory_hostname + '/' + inventory_hostname + '.crt') }}"
+    _localhost_cert_key_path: "{{ scylla_ssl.localhost_cert_key_path | default(inventory_dir + '/ssl/' + inventory_hostname + '/' + inventory_hostname + '.pem') }}"
```
tarzanek, line 5: why do you define them above if they get overriden anyways?

igorribeiroduarte:
> so with wildcard certificate support I guess #151 could be closed too

The issue above is still relevant when we are using certificates generated by this role, but it's out of the scope of the current PR

ansible-scylla-node/tasks/ssl.yml
```diff
@@ -35,11 +37,11 @@
   block:
     - name: Check if CA crt file exists
       stat:
-        path: "{{ _localhost_cert_path }}/ssl/ca/{{ scylla_cluster_name }}-ca.crt"
+        path: "{{ inventory_dir }}/ssl/ca/{{ scylla_cluster_name }}-ca.crt"
```
tarzanek, line 40: I guess you wanted a way to provide a custom dir and not to hardcode it ...

tarzanek: or wait, just CA has to be in same directory as inventory
and node certs can be elsewhere ?

tarzanek (approved):

tarzanek (approved):
thnx for the truststore resolution!

PR #437, opened 2025-03-17
scylla-node/templates: properly handle structured values

vladzcloudius:
This PR ensures that the JINJA templates output structured YAML values for scylla_manager_agent_config and io_properties instead of defaulting to JSON.

* Updates the io_properties comment to remove the pipe indicating a literal block style.
* Revises the scylla_manager_agent_config comment to clarify that the value is now expected as YAML.

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -414,18 +414,18 @@ scylla_manager_agent_upgrade: true
 
 # Manager Agent Backup configuration - this is a YAML value that will be inserted into the configuration file.
 # Please refer to the example configuration file and the documentation for detailed instructions.
-scylla_manager_agent_config:
-  s3:
-    access_key_id: 12345678
-    secret_access_key: QWerty123456789
-    provider: AWS
-    region: us-east-1c
-    endpoint: https://foo.bar.baz
-    server_side_encryption:
-    sse_kms_key_id:
-    upload_concurrency: 2
-    chunk_size: 50M
-    use_accelerate_endpoint: false
+#scylla_manager_agent_config:
+#  s3:
+#    access_key_id: 12345678
+#    secret_access_key: QWerty123456789
+#    provider: AWS
+#    region: us-east-1c
+#    endpoint: https://foo.bar.baz
+#    server_side_encryption:
+#    sse_kms_key_id:
+#    upload_concurrency: 2
+#    chunk_size: 50M
+#    use_accelerate_endpoint: false
 
 # Configure raid disks via scylla_setup. This requires a list of disks to add to the raid
 # scylla_raid_setup:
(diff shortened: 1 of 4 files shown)


ansible-scylla-node/defaults/main.yml
```diff
@@ -412,10 +412,9 @@ scylla_manager_rpm_repo_url: "{{ scylla_manager_repo_url }}"
...
 scylla_manager_agent_upgrade: true
 
-# Manager Agent Backup configuration - this is a free text which will be inserted into the configuration file
-# Please ensure it is a valid and working configuration. Please refer to the example configuration file and the
-# documentation for detailed instructions
-scylla_manager_agent_config: |
+# Manager Agent Backup configuration - this is a YAML value that will be inserted into the configuration file.
+# Please refer to the example configuration file and the documentation for detailed instructions.
+scylla_manager_agent_config:
   s3:
```
vreniers, line 418: Why is there even such a non-functional default set? This will require explicit override in practice. 
Hence this default should either be removed or placed in comments.

ansible-scylla-node/templates/io_properties.yaml.j2
```diff
@@ -1,5 +1,5 @@
 {% if scylla_io_probe|bool and scylla_io_probe_dc_aware|bool %}
-{{ hostvars[dc_to_node_list[dc][0]]['io_properties'] }}
+{{ hostvars[dc_to_node_list[dc][0]]['io_properties'] | to_yaml }}
 {% else %}
-{{ io_properties }}
+{{ io_properties | to_yaml}}
```
vreniers, line 4: In the current code this won't work entirely correctly. There is a task in ansible-scylla-node, namely `common.yml`, which may read out the IO properties from a file. Hence, this can still be a string.

```yml
- set_fact:
   io_properties: "{{ _io_properties.stdout }}"
```

The resulting file will for example look like:

```
cat io_properties.yaml

"disks:\n  - mountpoint: /var/lib/scylla\n    read_iops: 180008\n    read_bandwidth:
  737360000\n    write_iops: 100019\n    write_bandwidth: 409661376"
```

vreniers (changes requested):
Some comments.

ansible-scylla-node/defaults/main.yml
vladzcloudius: Good point. Let's comment it out. It was meant to be a demo - not a real default.

ansible-scylla-node/templates/io_properties.yaml.j2
vladzcloudius: fixd

ansible-scylla-node/defaults/main.yml
vladzcloudius: fixed

vreniers (approved):
Looks good.

vladzcloudius:
New in v3:
* Use `to_yaml(indent=2)` for a prettier output.


New in v2:
* Commented out a default `scylla_manager_agent_config` value in the Role.
* Fixed the parsing of `io_properties.yaml` file in the Role.

vreniers (approved):

PR #440, opened 2025-03-25
ansible-scylla-node: remove default scylla_seeds

vreniers:
The assumption cannot be made that the hostname is DNS resolvable by all nodes in the cluster. It's more appropriate to require that the seed nodes are specified explicitly.

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -295,9 +295,14 @@ io_conf: 'SEASTAR_IO="--io-properties-file /etc/scylla.d/io_properties.yaml"'
 #
 # scylla_args_value: "--log-to-syslog 1 --log-to-stdout 0 --default-log-level info --network-stack posix"
 
-# Seeds node list
-scylla_seeds:
-  - "{{ groups['scylla'][0] }}"
+# Seeds node list (mandatory parameter)
+# A seed should be either a DNS resolvable address (e.g. hostname) that is resolvable by all nodes in the cluster, 
+# or an IP address that is accessible by all nodes in the cluster.
+# If the hostname is not resolvable by all nodes, then such an IP address can be set, for example, explicitly or via ansible_host.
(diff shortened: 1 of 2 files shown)


ansible-scylla-node/defaults/main.yml
```diff
@@ -255,9 +255,14 @@ io_conf: 'SEASTAR_IO="--io-properties-file /etc/scylla.d/io_properties.yaml"'
...
-scylla_seeds:
-  - "{{ groups['scylla'][0] }}"
+# Seeds node list (mandatory parameter)
+# A seed should be either a DNS resolvable address (e.g. hostname) that is resolvable by all nodes in the cluster, 
+# or an IP address that is accessible by all nodes in the cluster.
+# If the hostname is not resolvable by all nodes, then such an IP address can be set, for example, explicitly or via ansible_host.
+# Do not use the below default example if ansible_host is not specified and the hostname is not resolvable by all nodes.
+#
+#scylla_seeds:
+#  - "{{ hostvars[groups.scylla.0].ansible_host if 'ansible_host' in hostvars[groups.scylla.0] else groups.scylla.0 }}"
```
vladzcloudius, line 265: If we remove the default then we must either add "if defined" blocks everywhere we use it or add an explicit check that this value is specified early in the play.

vladzcloudius (changes requested):

ansible-scylla-node/defaults/main.yml
vladzcloudius: ping, @vreniers

ManjotS:
@vreniers please address comments

ManjotS:
@vreniers please address comments.

ansible-scylla-node/defaults/main.yml
vreniers: Explicit check added.

vreniers:
> @vreniers please address comments.

Comments addressed.

ManjotS (approved):
Wait for Igor's approval

igorribeiroduarte (approved):

PR #449, opened 2025-04-08
Manager improvements: Improved defaults and allow override of Manager its local ScyllaDB version

vreniers:
This PR introduces a number of improvements:

- The deprecated parameter `scylla_manager_repo_url` has been removed.
- The manager repo URLs have been updated to a more recent version.
- Manager role now can re-use certain defaults present in the scylla-node role which were being overridden
- The Manager its local ScyllaDB version can now be specified instead of always being the latest version.
- Certain outdated defaults in the Manager playbook have been removed.

### DIFF
--- a/ansible-scylla-manager/defaults/main.yml
+++ b/ansible-scylla-manager/defaults/main.yml
@@ -2,11 +2,8 @@
 # defaults file for ansible-scylla-manager
 
 # Repo URL for the Manager
-# deprecated in favour of below, it will be dropped soon, below should be used
-scylla_manager_repo_url: "https://downloads.scylladb.com/rpm/centos/scylladb-manager-2.6.repo"
-
 scylla_manager_deb_repo_url: "https://downloads.scylladb.com/deb/ubuntu/scylladb-manager-2.6.list"
-scylla_manager_rpm_repo_url: "{{ scylla_manager_repo_url }}"
+scylla_manager_rpm_repo_url: "https://downloads.scylladb.com/rpm/centos/scylladb-manager-2.6.repo"
 
 scylla_manager_db_vars:
   # Repo URLs for the ScyllaDB datastore installation
--- a/ansible-scylla-manager/defaults/main.yml
+++ b/ansible-scylla-manager/defaults/main.yml
@@ -5,18 +5,17 @@
 scylla_manager_deb_repo_url: "https://downloads.scylladb.com/deb/ubuntu/scylladb-manager-3.5.list"
 scylla_manager_rpm_repo_url: "https://downloads.scylladb.com/rpm/centos/scylladb-manager-3.5.repo"
 
-scylla_manager_db_vars:
-  # Repo URLs for the ScyllaDB datastore installation
-  # More information on the variables can be found in the documentation for the scylla-node role and it's heavily
-  # commented `defaults/main.yml` file
-  # scylla_repos:
-  #   - 'https://repositories.scylladb.com/scylla/repo/centos/scylladb-4.6.repo'
+# The manager installs a local ScyllaDB, which is used only by the Manager.
+# Repo URLs for the ScyllaDB datastore installation
+# More information on the variables can be found in the documentation for the scylla-node role and it's heavily
+# commented `defaults/main.yml` file
+# scylla_repos:
+#   - 'https://repositories.scylladb.com/scylla/repo/centos/scylladb-4.6.repo'
 
-  # # Set when relevant (Debian/Ubuntu)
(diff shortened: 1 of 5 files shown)


ansible-scylla-manager/defaults/main.yml
```diff
@@ -2,27 +2,29 @@
...
+# scylla_repo_keys:
+#   - 5e08fbd8b5d6ec9c
+scylla_version: 'latest'
 scylla_repo_keyring_dir: /etc/apt/keyrings/
 scylla_repo_keyringfile: "{{ scylla_repo_keyring_dir }}/scylladb.gpg"
 
+scylla_manager_repo_keys:
+- d0a112e067426ab2
+- 5e08fbd8b5d6ec9c
+- a43e06657bac99e3
```
igorribeiroduarte, line 26: This is used by the node role to install the manager agent, not by the manager role
Why is it being moved to the manager role?

vreniers: - Manager uses the node role.
- It is not being moved. It is being copied. 
 
It ensures possible alignment between the current Manager version, and the potential use of a Manager agent if so desired on its local ScyllaDB. However, we can of course remove this. Do you suggest to remove it?

ansible-scylla-manager/defaults/main.yml
```diff
@@ -2,8 +2,8 @@
 # defaults file for ansible-scylla-manager
 
 # Repo URL for the Manager
-scylla_manager_deb_repo_url: "https://downloads.scylladb.com/deb/ubuntu/scylladb-manager-2.6.list"
-scylla_manager_rpm_repo_url: "https://downloads.scylladb.com/rpm/centos/scylladb-manager-2.6.repo"
+scylla_manager_deb_repo_url: "https://downloads.scylladb.com/deb/ubuntu/scylladb-manager-3.5.list"
+scylla_manager_rpm_repo_url: "https://downloads.scylladb.com/rpm/centos/scylladb-manager-3.5.repo"
```
vladzcloudius, line 6: They still don't have a `get latest` contact point like Scylla has (https://repositories.scylladb.com/scylla/check_version), do they?

igorribeiroduarte (changes requested):
The changes in this PR would prevent the usage of the two roles in a same playbook, since it would be impossible to set a different `scylla_version`/`scylla_edition` for each one of the two roles.
Having that in mind, I think it's better to keep `scylla_manager_db_vars`. This way we don't mix the usage of a same variable in two different roles.

ansible-scylla-manager/defaults/main.yml
vladzcloudius: We set `scylla_manager_enabled: false` for the internal Scylla, so these should not be needed. 
Let's remove them.

We can always add them if we need them in the future.

vladzcloudius (changes requested):
LGTM which a minor comment.
Please, address and let's merge.

vladzcloudius:
> The changes in this PR would prevent the usage of the two roles in a same playbook, since it would be impossible to set a different `scylla_version`/`scylla_edition` for each one of the two roles. Having that in mind, I think it's better to keep `scylla_manager_db_vars`. This way we don't mix the usage of a same variable in two different roles.

In what circumstances we would use both Roles in the same playbook, @igorribeiroduarte ?
Keeping `scylla_manager_db_vars` doesn't sound feasible to me because it will require a constant code syncing of all possible parameters we would want to "override" in the Manager internal Node Role's usage.

ansible-scylla-node/molecule/ubuntu2004-enterprise/molecule.yml
```diff
@@ -57,7 +57,7 @@ provisioner:
     group_vars:
       scylla:
         scylla_edition: "enterprise"
-        scylla_version: "latest"
+        scylla_version: "2024.2.1"
```
vladzcloudius, line 60: Please, let's use `2024.1.15` instead.

vladzcloudius (changes requested):

igorribeiroduarte:
> Keeping `scylla_manager_db_vars` doesn't sound feasible to me because it will require a constant code syncing of all possible parameters we would want to "override" in the Manager internal Node Role's usage.

Yeah, that's true. After thinking a little more I agree that the approach in this PR is better.

ansible-scylla-node/molecule/ubuntu2004-enterprise/molecule.yml
vreniers: Done.

igorribeiroduarte (approved):

vladzcloudius (approved):

PR #455, opened 2025-05-04
Monitoring role add scylla cluster name for alerts

vladzcloudius:
This pull request introduces several changes to enhance compatibility with newer Ubuntu versions, improve configurability of alerting, and ensure proper validation of required variables. The most important changes include adding support for Ubuntu 22.04 in the Molecule testing framework, introducing a new variable for alert customization, and improving validation and default handling for cluster name variables.

### Ubuntu Compatibility and Testing Enhancements:
* Added Ubuntu 22.04 to the test matrix in `.github/workflows/ansible-scylla-monitoring.yml`. Ubuntu 24.04 is commented out due to issues with fetching GPG keys.
* Introduced new Molecule configuration files for Ubuntu 22.04, including `molecule.yml` and a basic `verify.yml` playbook. [[1]](diffhunk://#diff-a26c3a36ab5ec54afebb0b6511d3336be93da4685cd0162f2221e0e2e9dafd3eR1-R40) [[2]](diffhunk://#diff-719561619e3d604edc4f82d1609e30166b4e4ecdac111a4668bbe64764e6cee4R1-R9)

### Alerting Configuration Improvements:
* Added a new variable `scylla_cluster_name_for_alerts` to allow customization of the cluster name used in alert subjects. This defaults to the value of `scylla_cluster_name` if not explicitly set. [[1]](diffhunk://#diff-610bf15fd475a2628bc1cb5d3090a5a98db4d7535854f8d001210c1c183aff69R4-R6) [[2]](diffhunk://#diff-d910f08594d27686bb2fe4cc7fa438901009dbb41d7fd524d3e84f947d82b440R214-R218)
* Updated alert email subject templates in `rule_config.yml.j2` to use `scylla_cluster_name_for_alerts` instead of `scylla_cluster_name`. [[1]](diffhunk://#diff-386e61f238d04e6e5839ccdd582da4dd18dfec4fefa309ff9e51cf7d23cbc468L53-R53) [[2]](diffhunk://#diff-386e61f238d04e6e5839ccdd582da4dd18dfec4fefa309ff9e51cf7d23cbc468L64-R64)

### Validation and Default Handling:
* Added a task in `tasks/main.yml` to fail if `scylla_cluster_name` is not defined, ensuring this critical variable is always set.
* Set a default value for `scylla_cluster_name_for_alerts` in `tasks/common.yml` if it is not explicitly defined.

These changes collectively improve the robustness, flexibility, and compatibility of the `ansible-scylla-monitoring` project.


ansible-scylla-monitoring/molecule/ubuntu2204/verify.yml
vreniers, line 1: What's the purpose of this file? It seems like it does nothing.

ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -211,6 +211,11 @@
     mode: '0644'
   when: rule_config_stat.stat.exists|bool
 
+- name: Set the default scylla_cluster_name_for_alerts if not defined
+  set_fact:
+      scylla_cluster_name_for_alerts: "{{ scylla_cluster_name }}"
+  when: scylla_cluster_name_for_alerts is not defined
```
vreniers, line 217: Why not just define this among the other defaults? It's a bit of an inconsistent approach (cf. other comment).

ansible-scylla-monitoring/defaults/main.yml
```diff
@@ -1,6 +1,9 @@
 ---
 # scylla_cluster_name: 'testcluster123'
 
+# This is the name the cluster is going to reported by the Alert Manager. This defaults to a scylla_cluster_name value.
+# scylla_cluster_name_for_alerts: 'testcluster123'
```
vreniers, line 5: This should become:

```yml
scylla_cluster_name_for_alerts: "{{ scylla_cluster_name_for_alerts | default(scylla_cluster_name) }}"
```

ansible-scylla-monitoring/templates/rule_config.yml.j2
```diff
@@ -50,7 +50,7 @@ receivers:
     auth_password: {{ alerts_sender_password }}
     send_resolved: false
     headers:
-      Subject: '[PRODUCTION] {{ '{{ .CommonLabels.alertname }}' }} {{ '{{ .CommonLabels.severity }}' }}: NORMAL Managed-Services - Cluster:{{ scylla_cluster_name }}'
+      Subject: '[PRODUCTION] {{ scylla_cluster_name_for_alerts }}:{{ '{{ .CommonLabels.alertname }}' }} {{ '{{ .CommonLabels.severity }}' }}: NORMAL'
```
vreniers, line 53: Out-of-scope, but regarding this "`[PRODUCTION]`", it seems this shouldn't be hardcoded either.

.github/workflows/ansible-scylla-monitoring.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       matrix:
         molecule_distro:
           - ubuntu2004
+          - ubuntu2204
+          #- ubuntu2404 - there is some weird issue with fetching apa gpg keys. Let's disable this target for now
```
vreniers, line 22: If we're disabling it right now, does it make sense to add it then already?

vreniers (changes requested):
Some minor comments, overall looks good.

ansible-scylla-monitoring/defaults/main.yml
vladzcloudius: I'm not sure we can assume that `scylla_cluster_name` would be visible at this point, can we?

ansible-scylla-monitoring/templates/rule_config.yml.j2
vladzcloudius: I agree. I suggest dropping it completely.

.github/workflows/ansible-scylla-monitoring.yml
vladzcloudius: It's a draft - not PR yet. I'll try to fix it by the time it's mergable.

ansible-scylla-monitoring/tasks/common.yml
vladzcloudius: Agree

ansible-scylla-monitoring/defaults/main.yml
vladzcloudius: done

ansible-scylla-monitoring/tasks/common.yml
vladzcloudius: done

.github/workflows/ansible-scylla-monitoring.yml
vladzcloudius: Dropping it for now - will try to fix this in a follow up. 
It's probably related to the fact that we are using a U20.04 runner - will try to use U24.04 instead.

ansible-scylla-monitoring/templates/rule_config.yml.j2
vladzcloudius: done

ansible-scylla-monitoring/molecule/ubuntu2004/verify.yml
vreniers, line 1: If you remove this, then I think this should also be removed from the molecule its test sequence, because now the following happens:

```
Molecule ubuntu2004 > verify
  INFO     Running Ansible Verifier
  WARNING  Skipping, verify action has no playbook.
  INFO     Verifier completed successfully.
```

vreniers (changes requested):
Looks good, thank you. One last comment.

ansible-scylla-monitoring/molecule/ubuntu2004/verify.yml
vladzcloudius: I'll reference the one from the `shared` instead - so it's in front of us in case we need to put something there.

vladzcloudius: done

vreniers (approved):
Looks good. Thank you.

PR #457, opened 2025-05-13
Overall improvements and increase apt lock timeout

vreniers:
This PR introduces some general improvements, with the main purpose to introduce a module default for the apt package that has an increased timeout. We observed in some cases that another process (e.g. automatic software update) can claim a long apt lock, this can cause automation fail. In order to prevent this in most cases, we introduce a default for the apt module with a sufficiently long timeout.

Such a module default can only be introduced at the playbook, block, or task level. In order to set it, we introduce some general improvements to mainly make the main logic of each role more clear, and cohesive. 
Subsequently, we introduce a block in each `main.yml` that has appropriate defaults for the apt module.

Overview of changes:
- scylla-manager: Rename `add-clusters.yml` to `add_cluster.yml` as per proper naming convention and to more accurately reflect its logic.
- scylla-manager: Group logic into `start_manager.yml` task file..
- scylla-node: Set `scylla_package_prefix` in the `defaults.yml` and not in the `main.yml`
- scylla-node: Group logic regarding version check into a separate file
- scylla-monitoring: Remove a redundant check on `scylla_cluster_name`
- wipe-node.yml playbook: Set apt lock timeout on the apt task
- scylla-node, scylla-manager, scylla-monitoring: Set default apt lock timeout


ansible-scylla-node/defaults/main.yml
```diff
@@ -208,6 +208,9 @@ scylla_version: '0.0.0'
 # Options are oss|enterprise
 #scylla_edition: oss
 scylla_edition: enterprise
+scylla_package_prefix: >-
+  {{ 'scylla-enterprise' if scylla_edition == 'enterprise' else
+     'scylla' }}
```
vladzcloudius, line 213: Why this strange syntax?
Wouldn't you achieve the same with:

```
scylla_package_prefix: "{{ 'scylla-enterprise' if scylla_edition == 'enterprise' else 'scylla }}"
```

?

example-playbooks/wipe_node/Debian.yml
```diff
@@ -4,3 +4,4 @@
     name: "scylla*"
     state: absent
     purge: true
+    lock_timeout: 300
```
vladzcloudius, line 7: I don't think 5 minutes is enough. Let's make it 20.

ansible-scylla-manager/defaults/main.yml
```diff
@@ -52,3 +52,5 @@ enable_upgrade: false
 #     without_repair: false
 #     host: 1.2.3.4
 
+# Default time-out to wait on acquiring an APT lock.
+apt_lock_timeout: 300
```
vladzcloudius, line 56: Let's make it 20 minutes. This value should be a "reasonable infinity" and 5 minutes is not long enough IMO. A kernel upgrade on an overloaded system can take 5 minutes.

ansible-scylla-monitoring/defaults/main.yml
```diff
@@ -98,6 +98,9 @@ prometheus_url: 'https://github.com/prometheus/prometheus/releases/download/v2.3
 # The content of this variable will be coppied to /etc/docker/daemon.json
 # docker_daemon_custom_config: { "bip": "172.17.0.1/24" }
 
+# Default time-out to wait on acquiring an APT lock.
+apt_lock_timeout: 300
```
vladzcloudius, line 102: Same comment here - let's make it 1200.

vladzcloudius (changes requested):
Looks good in general.
Had a few comments however.

ansible-scylla-monitoring/tasks/main.yml
```diff
@@ -1,9 +1,4 @@
 ---
-- name: Verify that a scylla_cluster_name is set
-  fail:
-    msg: "scylla_cluster_name must be defined"
-  when: scylla_cluster_name is not defined
```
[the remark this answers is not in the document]
vreniers: I don't think this check may be reached at all in its current state because the defaults.yml will be reached first.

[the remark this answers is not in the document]
vladzcloudius: No, Ansible doesn't work like that ;)
Defaults (or any variables in general) are not resolved until they are used.

ansible-scylla-node/defaults/main.yml
vreniers: It was a bit longer originally, hence the reason for this syntax as I had thought about also including a version check in there (if version  > 2025) then default to 'scylla' either way.

ansible-scylla-monitoring/defaults/main.yml
vreniers: Done.

ansible-scylla-manager/defaults/main.yml
vreniers: Done.

example-playbooks/wipe_node/Debian.yml
vreniers: Done.

ansible-scylla-node/defaults/main.yml
vreniers: Adjusted to a single liner.

ansible-scylla-monitoring/tasks/main.yml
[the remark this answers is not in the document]
vreniers: You're right. Dropped this change.

PR #468, opened 2025-06-30
Validate ports

igorribeiroduarte:


ansible-scylla-common/defaults/main.yml
```diff
@@ -3,4 +3,13 @@
...
+disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
+
+# A dictionary defining which TCP ports should be reachable on the node running this role,
+# keyed by the name of the source group.
+# For each group, the role will attempt to establish a TCP connection
+# from each host in that group to the ports listed.
+# required_ports:
+#   scylla: []
+#   scylla-manager: []
+#   scylla-monitor: []
```
vreniers, line 15: Generally speaking, you want to avoid leaking domain concepts into a common role. There's for example no need to specify aspects about `scylla`, `scylla-manager`, etc. There's for example also possibly the loader group, so the question is then why it's omitted. Or perhaps other groups will come, or the names will change.

It will promote extensibility, cohesion, and maintainability. In fact, this is also what this functionality allows to do today. So it's more of a documentation issue than anything else. E.g.

```
# required_ports:
#     group_1: [ ]
#     group_2: [ ]
```

Although, concretizing the example is definitely also a nice touch from practical point of view. But I would start generic.

ansible-scylla-common/defaults/main.yml
```diff
@@ -3,4 +3,13 @@
...
 # Disable firewall (default is true).
 # When the deprecated variable `firewall_enabled` is present, and when this is false, the firewall will be deactivated.
-disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
\ No newline at end of file
+disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
+
+# A dictionary defining which TCP ports should be reachable on the node running this role,
+# keyed by the name of the source group.
+# For each group, the role will attempt to establish a TCP connection
+# from each host in that group to the ports listed.
```
vreniers, line 11: Clarify: Only connectivity is checked from each host in the group, to the current host its listed ports.

Perhaps change in variable name could also help with this.

`required_ports_open_to`

ansible-scylla-common/tasks/check_ports.yml
```diff
@@ -0,0 +1,41 @@
...
+  poll: 0
+
+- name: Check if all the necessary ports are open
+  shell:
+    "nc -z -w 2 {{ hostvars[inventory_hostname]['ansible_host'] | default(inventory_hostname) }} {{ item[1] }}"
+  when: item[0] != inventory_hostname
+  delegate_to: "{{ item[0] }}"
+  with_nested:
+    - "{{ groups[source_host_group] | default([]) }}"
+    - "{{ dest_ports }}"
```
vreniers, line 35: There's documentation on Ansible suggesting to migrate to `loop`.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#with-nested-with-cartesian

ansible-scylla-common/tasks/check_ports.yml
```diff
@@ -0,0 +1,41 @@
...
+      - iproute2
+    state: present
+  become: true
+
+- name: Create temporary listeners on required ports, if needed
+  shell: |
+    if ! ss -ltn | awk '{print $4}' | grep -q ":{{ port }}$"; then
+      timeout 60 socat TCP-LISTEN:{{ port }},fork EXEC:'/bin/true'
+    fi
+  loop: "{{ dest_ports }}"
```
vreniers, line 22: Nitpick variable naming: The `dest_ports` is kind of relative, because here it is just a local port.

ansible-scylla-common/tasks/check_ports.yml
```diff
@@ -0,0 +1,41 @@
...
+      - socat
+      - netcat
+      - iproute2
+    state: present
+  become: true
+
+- name: Create temporary listeners on required ports, if needed
+  shell: |
+    if ! ss -ltn | awk '{print $4}' | grep -q ":{{ port }}$"; then
+      timeout 60 socat TCP-LISTEN:{{ port }},fork EXEC:'/bin/true'
```
vreniers, line 20: Is this required?

vreniers (changes requested):
Hi @igorribeiroduarte, nice work. I really like the fact that you placed it inside of the common-role, it's already a nice contribution to this new role. 

I have some comments, most are just clarification in naming and documentation, and some slight improvements.

ansible-scylla-common/tasks/check_ports.yml
igorribeiroduarte: Done

ansible-scylla-common/defaults/main.yml
igorribeiroduarte: Fixed

ansible-scylla-common/tasks/check_ports.yml
```diff
@@ -0,0 +1,41 @@
...
+    fi
+  loop: "{{ dest_ports }}"
+  loop_control:
+    loop_var: port
+  async: 70
+  poll: 0
+
+- name: Check if all the necessary ports are open
```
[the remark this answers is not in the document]
igorribeiroduarte: Adjusted

ansible-scylla-common/defaults/main.yml
```diff
@@ -3,4 +3,13 @@
 
 # Disable firewall (default is true).
 # When the deprecated variable `firewall_enabled` is present, and when this is false, the firewall will be deactivated.
-disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
\ No newline at end of file
+disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
+
+# A dictionary defining which TCP ports should be reachable on the node running this role,
```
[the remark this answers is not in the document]
igorribeiroduarte: Adjusted to:
> reachable on the node(s) targeted by this role

ansible-scylla-common/tasks/check_ports.yml
igorribeiroduarte: Since the `common` role is executed before installing `scylla` and `scylla-manager`, if this is the first time running these roles, no one will be listening on these ports and in such case we need to create a temporary listener just to be able to test these ports.

ansible-scylla-common/tasks/check_ports.yml
igorribeiroduarte: Renamed to `ports_to_test`

ansible-scylla-common/defaults/main.yml
igorribeiroduarte: Adjusted

PR #469, opened 2025-07-07
ansible-scylla-monitoring: Add support for node alias

igorribeiroduarte:


ansible-scylla-monitoring/tasks/common.yml
```diff
@@ -107,16 +107,28 @@
...
-    # Resulting map -> {'-dc dc1': [node1, node2, node3], '-dc dc2': [node4, node5, node6]}
+    - name: Extract monitoring version
+      set_fact:
+        _monitoring_version: "{{ scylla_monitoring_archive_url | regex_search('([0-9]+(?:\\.[0-9]+)*)\\.tar\\.gz$', '\\1') | first }}"
+
+    - name: Initialize the dc_to_node_list map
+      set_fact:
+        dc_to_node_list: {}
+
+    # Resulting map -> {'-dc dc1': [node1:label1, node2:label2, node3:label3], '-dc dc2': [node4:label4, node5:label5, node6:label6]}
```
vreniers, line 118: This documentation should be adjusted. The '-dc' does not get added here. 
Therefore this is not the resulting map.

vreniers (changes requested):
A small comment on documentation, otherwise looks good.

I tested it also on my cluster, and it worked nicely.

ansible-scylla-monitoring/tasks/common.yml
igorribeiroduarte: Fixed

vreniers (approved):

PR #488, opened 2025-08-29
check_ports.yml: Parametrize and increase default client timeout

igorribeiroduarte:


vladzcloudius:
@vreniers increasing the timeout alone is not going to fix the issue.
The problem was never the value of the timeout - 2 seconds is more than enough for any network on planet Earth. 
The problem that you were hitting was very likely caused by the fact that the listener was not actually listening when the client was started. And if this was the case, increasing the timeout is not fixing anything - we will hit this again when if it takes the server to start long enough.

vreniers:
> @vreniers increasing the timeout alone is not going to fix the issue. The problem was never the value of the timeout - 2 seconds is more than enough for any network on planet Earth. The problem that you were hitting was very likely caused by the fact that the listener was not actually listening when the client was started. And if this was the case, increasing the timeout is not fixing anything - we will hit this again when if it takes the server to start long enough.

@vladzcloudius Please make a relevant issue.

PR #495, opened 2025-09-19
generate_certificate_and_key.yml: Generate a new certificate only if a local one doesn't exist

igorribeiroduarte:


vreniers (approved):
Thx @igorribeiroduarte. Good work.

PR #510, opened 2026-03-31
ansible-scylla-common: dpkg set timeout using become flag

vreniers:
We need root privileges in order to be able to write the apt configuration.


vladzcloudius (changes requested):
@vreniers , please, add a another commit to this PR that reverts https://github.com/scylladb/scylla-ansible-roles/pull/511 (which was required due to a bug this PR fixes).

vladzcloudius (approved):

PR #520, opened 2026-05-12
 ansible-scylla-common: Replace port lists with ip/port tuples in `required_ports_open_to`

igorribeiroduarte:
This pull request updates the way required TCP ports are specified and checked in the Ansible roles for Scylla, moving from a port-only approach to an explicit IP and port structure. It also refactors how common configuration variables are managed and passed between roles, centralizing and clarifying their usage. The changes improve flexibility and clarity for connectivity checks and configuration management.

**Connectivity Check Improvements:**

* Changed the `required_ports_open_to` variable from a list of ports to a list of dictionaries specifying both `ip` and `port`, allowing more granular and explicit connectivity requirements.
* Updated the port listener and connectivity check tasks in `check_ports.yml` to use both `ip` and `port`, ensuring listeners bind to the correct address and connectivity is validated per IP/port pair.

**Configuration Management Refactoring:**

* Removed default definitions of `apt_lock_timeout`, `repo_pattern_list`, and `required_ports_open_to` from individual role defaults (`ansible-scylla-manager`, `ansible-scylla-monitoring`, `ansible-scylla-node`) and moved their assignment to the role dependencies in `meta/main.yml`, centralizing configuration and reducing duplication. [[1]](diffhunk://#diff-24344c2842b4ea1bdbf6c60a2a6043a71eca4b9f6810081738d2dd78aaf07ad9L54-L62) [[2]](diffhunk://#diff-610bf15fd475a2628bc1cb5d3090a5a98db4d7535854f8d001210c1c183aff69L101-L106) [[3]](diffhunk://#diff-20829e5f73bbd6db3dff12e61bd63ef7f7b3daa0b647bb4b0a34b0dbb05ba6c6L602-L611)
* Explicitly set role variables for dependencies in the `meta/main.yml` files for `ansible-scylla-manager`, `ansible-scylla-monitoring`, and `ansible-scylla-node`, ensuring that common configuration is consistently and correctly passed to `ansible-scylla-common`. [[1]](diffhunk://#diff-8c5a41795ce359e665fa5201923a6753fd9f92a67f4a7dcf2f7456d738a3fbecR29-R42) [[2]](diffhunk://#diff-780bd9228577abc9a4da89a1b40e60297c527cb89c017e93ceff4f344b744d6bR31-R35) [[3]](diffhunk://#diff-1791552ae8392bf97da2377fd3bab65917430fe5dcd760264a3d50c3f5549ef7R33-R46)

### DIFF
--- a/ansible-scylla-common/defaults/main.yml
+++ b/ansible-scylla-common/defaults/main.yml
@@ -5,13 +5,19 @@
 # When the deprecated variable `firewall_enabled` is present, and when this is false, the firewall will be deactivated.
 disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
 
-# A dictionary that defines which TCP ports must be reachable on the node(s) targeted by this role.
+# A dictionary that defines which TCP addresses (IP or DNS) and ports must be reachable on the node(s) targeted by this role.
 # The dictionary is keyed by the name of the source group.
 # For each source group, the role will test connectivity by attempting to establish a TCP connection
-# from each host in the group to the current host on the listed ports.
+# from each host in the group to the current host on the list.
 # required_ports_open_to:
-#   group_1: []
-#   group_2: []
+#   group_1:
+#     - address: 1.2.3.4
+#       port: 8000
+#   group_2:
+#     - address: 1.2.3.5
+#       port: 8001
+#     - address: 1.2.3.5
+#       port: 8002
 #
 # Default duration that listeners used for testing required ports should stay alive
 check_ports_listener_timeout_sec: 360
--- a/ansible-scylla-common/tasks/check_ports.yml
+++ b/ansible-scylla-common/tasks/check_ports.yml
@@ -13,9 +13,7 @@
 
 - name: Create temporary listeners on required ports, if needed
   shell: |
-    if ! ss -ltn | awk '{print $4}' | grep -q "^{{ address }}:{{ port }}$"; then
-      timeout {{ check_ports_listener_timeout_sec | int }} socat TCP-LISTEN:{{ port }},bind={{ address }},fork EXEC:'/bin/true'
-    fi
+    timeout {{ check_ports_listener_timeout_sec | int }} socat TCP-LISTEN:{{ port }},bind={{ address }},fork EXEC:'/bin/true'
   loop: "{{ ports_to_test }}"
   vars:
     address: "{{ item.address }}"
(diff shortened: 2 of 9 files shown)


vladzcloudius:
Please, take a look why CI/CD failed, @igorribeiroduarte

ansible-scylla-common/defaults/main.yml
```diff
@@ -5,13 +5,19 @@
 # When the deprecated variable `firewall_enabled` is present, and when this is false, the firewall will be deactivated.
 disable_firewall: "{{ not (firewall_enabled | default(false)) }}"
 
-# A dictionary that defines which TCP ports must be reachable on the node(s) targeted by this role.
+# A dictionary that defines which TCP IPs and ports must be reachable on the node(s) targeted by this role.
```
vladzcloudius: Does it have to be an IP? Can't it be a DNS name? I think it can be.

ansible-scylla-common/tasks/check_ports.yml
```diff
@@ -13,30 +13,30 @@
 
 - name: Create temporary listeners on required ports, if needed
   shell: |
-    if ! ss -ltn | awk '{print $4}' | grep -q ":{{ port }}$"; then
-      timeout {{ check_ports_listener_timeout_sec | int }} socat TCP-LISTEN:{{ port }},fork EXEC:'/bin/true'
+    if ! ss -ltn | awk '{print $4}' | grep -q "^${{ ip }}:{{ port }}$"; then
```
vladzcloudius: That's a strange grep expression.
Didn't you mean: `grep -q "^{{ ip }}:{{ port }}$"`

vladzcloudius (changes requested):
Looks good in general. A few relatively minor comments.

ansible-scylla-common/tasks/check_ports.yml
igorribeiroduarte: yes, that's exactly what I meant. I probably added the `$` by mistake during rebase

ansible-scylla-common/defaults/main.yml
igorribeiroduarte: Yes, it can. I've renamed `ip` to `address` and adjusted the description

ansible-scylla-common/tasks/check_ports.yml
vladzcloudius: This was supposed to break the corresponding test.
Which suggests that there isn't any test covering this feature.
Do we use it in the molecule? We should.

igorribeiroduarte: >  Do we use it in the molecule? We should.

We do, but this line is removed in the following patch ("ansible-scylla-common: Always attempt socat listener creation for port checks")
So it doesn't get tested

vladzcloudius (approved):

PR #522, opened 2026-06-20
ansible-scylla-monitoring: Update the Monitor deployment to support arm64 architecture

faizalrf:
This PR addresses the Scylla Monitor deployment failure on an ARM based VM.

### DIFF
--- a/ansible-scylla-monitoring/tasks/Debian.yml
+++ b/ansible-scylla-monitoring/tasks/Debian.yml
@@ -40,7 +40,7 @@
 
   - name: add repo
     ansible.builtin.apt_repository:
-      repo: "deb [arch=amd64] https://download.docker.com/linux/{{ ansible_distribution | lower }} {{ ansible_distribution_release }} stable"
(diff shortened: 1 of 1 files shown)


igorribeiroduarte (changes requested):
The PR looks good. Thanks for the contribution!
Could you please adjust the commit message to match the pattern used in this repository?

faizalrf:
@igorribeiroduarte I have made the required updates. I am wondering if other OS have the same issue though!

igorribeiroduarte:
> @igorribeiroduarte I have made the required updates. I am wondering if other OS have the same issue though!

I haven't tested it, but I wouldn't expect it to be an issue.
On Debian based systems we need to specify the architecture in the repository definition, but that isn't needed for RPM-based systems. If you look at the repo file being downloaded:
https://github.com/scylladb/scylla-ansible-roles/blob/master/ansible-scylla-monitoring/tasks/RedHat.yml#L51
https://download.docker.com/linux/centos/docker-ce.repo

you'll see it uses the `$basearch` variable, which is expanded automatically by `yum`/`dnf` to the system's architecture.

igorribeiroduarte (approved):

igorribeiroduarte (approved):
CI is failing for unrelated reasons that will be addressed in a separate PR.
LGTM.

PR #508, opened 2026-02-20
ansible-scylla-node: Validate that scylla-server is running while waiting for CQL port

igorribeiroduarte:
This pull request refactors the process for starting Scylla nodes in the Ansible playbooks, focusing on improved variable handling, enhanced startup checks, and more consistent timeout management. The changes primarily streamline how nodes are started and monitored, making the playbooks more robust and easier to maintain.

Improvements to node startup and monitoring:

* Replaced the use of `item` with a dedicated `node` variable when starting nodes, ensuring clearer and more consistent variable usage across tasks and roles (`ansible-scylla-node/tasks/common.yml`, `ansible-scylla-node/tasks/start_one_node.yml`, `example-playbooks/replace_node/replace_node.yml`). [[1]](diffhunk://#diff-b6c93256c8075ca686527b80578847d5c50e8c85a714f9a41851c610152126dcR325-R334) [[2]](diffhunk://#diff-39c31111da077f316221697afd406671a386a932c56c2cb1e6a20dd1c62ddc02L2-R30) [[3]](diffhunk://#diff-0669256df2a7ebd6796cee2327bcdfe0fe3fbcd3354f28bad82f7d09e1e69631L297-R314)
* Enhanced the CQL port check by switching from the `wait_for` module to a custom shell command that also verifies the Scylla server's state, providing better error handling and more reliable startup detection (`ansible-scylla-node/tasks/start_one_node.yml`).

Timeout and dependency management:

* Unified the bootstrap timeout by removing the `new_node_bootstrap_wait_time_sec` variable and switching to `scylla_bootstrap_wait_time_sec` for all relevant operations, ensuring consistency across node replacement and startup tasks (`example-playbooks/replace_node/replace_node.yml`, `example-playbooks/replace_node/vars/main.yml`). [[1]](diffhunk://#diff-0669256df2a7ebd6796cee2327bcdfe0fe3fbcd3354f28bad82f7d09e1e69631L358-R357) [[2]](diffhunk://#diff-5ce273d3a5fa9ea724b25634719bdad94dfffce713d4b86f61c9bdcc50f1ea53L30-L37)
[...]

PR #524, opened 2026-06-30
Fail fast when scylla-server stops during bootstrap health waits

ManjotS:
## Summary
- Check `systemctl is-active scylla-server` while polling nodetool during cluster health wait in the node role
- Apply the same fail-fast pattern to replace_node UN health wait (delegate to new node)
- Matches the existing CQL wait logic in `start_one_node.yml` (PR #508)

Fixes scylladb/field-engineering#2109

## Test plan
- [ ] Replace_node: simulate mid-bootstrap scylla crash; job fails immediately instead of waiting full timeout
- [ ] Create_cluster / Add_new_nodes: cluster health wait fails fast if a node dies during bootstrap
- [ ] Normal bootstrap still succeeds when scylla reaches UN

### DIFF
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -344,13 +344,28 @@
         timeout: 300
 
     - name: wait for the cluster to become healthy
-      shell: |
-        nodetool status|grep -E '^UN|^UJ|^DN'|wc -l
+      shell:
+        executable: /bin/bash
+        cmd: |
+          state=$(systemctl is-active scylla-server)
+
+          if [[ "$state" == "failed" || "$state" == "inactive" ]]; then
+            echo "scylla-server has unexpectedly stopped"
+            exit 1
+          fi
+
+          count=$(nodetool status | grep -E '^UN|^UJ|^DN' | wc -l)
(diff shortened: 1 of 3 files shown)


ManjotS:
Retriggered CI (empty commit).

The debian12 failure was **not** caused by this PR:
- `wait for the cluster to become healthy` passed on node1/2/3
- Failure was `scylla-manager-node` timing out waiting for `127.0.0.1:5080` (manager HTTP port)

Same molecule flake seen on master (e.g. rockylinux10). Re-running now that runners were backed up.

vreniers (approved):
It's a bit of a behavioral change for the replace job. Before the change, all nodes checked whether the replaced node appeared as `UN`. Now, we only inspect it locally.

This is comment more important for clusters with gossip-based topology. Versions after `2025.x` should be on Raft topology so I don't think this is a problem there. Since before 2025 are unsupported version, it's ok for me.

Furthermore, regarding node role, but not related to this PR, it's a bit strange to count `DN` as an element towards a healthy cluster? 

@ManjotS can we merge these into a `dev` branch, and then stage releases?

ManjotS:
@vreniers base is changed but our related PR does rely on this merge.

I am not concerned about versions below 2025.

PR #536, opened 2026-07-25
ansible-scylla-monitoring: optional systemd boot service

ManjotS:
## Summary
- Add opt-in `install_scylla_monitoring_service` (default `false`)
- When enabled, install and enable `scylla-monitoring.service` so the stack starts after reboot (and when `docker.service` restarts via `PartOf=`)
- Unit calls the existing start/stop wrapper scripts under `scylla_monitoring_deploy_path`

## Test plan
- [ ] Deploy monitoring with default vars — confirm no systemd unit installed
- [ ] Deploy with `-e install_scylla_monitoring_service=true` — confirm unit present, enabled, and `systemctl cat scylla-monitoring` points at wrapper scripts
- [ ] Reboot monitor host — stack comes back via systemd
- [ ] `systemctl restart docker` — monitoring stops/starts with Docker (`PartOf=docker.service`)


ManjotS (commented):
This adds an opt-in systemd unit (`install_scylla_monitoring_service`, default `false`) that installs and enables `scylla-monitoring.service`, wiring it to the existing start/stop wrapper scripts so the monitoring stack survives reboots and `docker.service` restarts via `PartOf=`. The change is small, additive, and off by default, so it carries low risk to existing deployments.

Overall this looks solid — the `Requires=`/`After=`/`PartOf=docker.service` combination on the unit is the correct, idiomatic way to tie a docker-compose-style service to `docker.service`'s lifecycle (see inline praise). A few things worth a look:

- The unit is `enable`d but never `start`ed by Ansible, so `systemctl status scylla-monitoring` will report `inactive (dead)` right after a deploy even though the stack is actually running — see inline comment for details.
- The new code path (`install_scylla_monitoring_service: true`) isn't exercised anywhere in the molecule suite, so CI won't catch regressions in the template or task logic.
- A small style nit (FQCN vs. the file's existing short module names) and a question about the disable/cleanup story when the flag is flipped back off.

Nothing here is blocking — the feature is off by default and functionally sound for the stated goal (reboot survival + docker-restart propagation).

**Coverage note:** reviewed correctness/logic, security, reliability/operability, maintainability, and test coverage for the diff. I read the role's task graph and templates but did not have a live host to actually exercise the reboot / `systemctl restart docker` test-plan items from the PR description — those are assessed from the unit file and task logic, not from running them.

ansible-scylla-monitoring/tasks/docker.yml
```diff
@@ -79,3 +79,21 @@
...
+    group: root
+    mode: "0644"
+  become: true
+  when: install_scylla_monitoring_service | bool
+
+- name: "docker.yml: Enable scylla-monitoring at boot"
+  ansible.builtin.systemd:
+    name: scylla-monitoring.service
+    enabled: true
+    daemon_reload: true
```
ManjotS: 🟠 **Important** — The unit is `enable`d here but never `state: started`, so right after a normal Ansible run `systemctl status scylla-monitoring.service` (and `is-active`) will report `inactive (dead)` — even though the stack is actually up, because it was started a few tasks earlier via the raw `command: {{ scylla_monitoring_start_script_path }}` task, outside systemd's tracking. That's a surprising status for anyone checking `systemctl status` right after a deploy; the discrepancy only resolves itself after the next reboot or `systemctl restart docker`.

```suggestion
    name: scylla-monitoring.service
    enabled: true
    state: started
    daemon_reload: true
```

This assumes it's safe for the start script to run twice in one playbook run (once via the `command` task above, once via `state: started` here) — worth confirming `start-all.sh` is idempotent for that, or consider letting systemd own start/stop entirely when this flag is set (drop the manual `command` tasks in that branch) so there's a single source of truth for the running state.

ansible-scylla-monitoring/tasks/docker.yml
```diff
@@ -79,3 +79,21 @@
...
+    src: scylla-monitoring.service.j2
+    dest: /etc/systemd/system/scylla-monitoring.service
+    owner: root
+    group: root
+    mode: "0644"
+  become: true
+  when: install_scylla_monitoring_service | bool
+
+- name: "docker.yml: Enable scylla-monitoring at boot"
+  ansible.builtin.systemd:
```
ManjotS: 🔵 nit: the rest of this file uses short module names (`template`, `service`, `command`, `file`, `set_fact`) rather than FQCNs. For consistency, `systemd:` would match the `service:` task above (lines 56-61) that does the equivalent docker-restart job.

ansible-scylla-monitoring/tasks/docker.yml
```diff
@@ -79,3 +79,21 @@
   environment:
     SCYLLA_USER: "{% if monitoring_cql_username is defined and monitoring_cql_password is defined %}{{ monitoring_cql_username }}{% endif %}"
     SCYLLA_PSSWD: "{% if monitoring_cql_username is defined and monitoring_cql_password is defined %}{{ monitoring_cql_password }}{% endif %}"
+
+- name: "docker.yml: Install scylla-monitoring systemd unit"
```
ManjotS: 🟡 Minor — `molecule/shared/converge.yml` includes this role but never sets `install_scylla_monitoring_service: true`, so this template-install/enable path (and the new `scylla-monitoring.service.j2` template) isn't exercised by CI at all. Since the flag defaults to `false`, existing coverage still passes, but a regression here (bad template syntax, wrong `when`, etc.) wouldn't be caught. Worth adding a scenario/host var that turns it on plus a `verify.yml` check for the installed and enabled unit.

ansible-scylla-monitoring/defaults/main.yml
```diff
@@ -27,6 +27,9 @@ scylla_monitoring_deploy_path: /opt/scylla-monitoring
 scylla_monitoring_start_script_path: "{{ scylla_monitoring_deploy_path }}/start-monitoring.sh"
 scylla_monitoring_stop_script_path: "{{ scylla_monitoring_deploy_path }}/stop-monitoring.sh"
 scylla_monitoring_start_env_path: "{{ scylla_monitoring_deploy_path }}/start-monitoring.env"
+# Install and enable a systemd unit that starts/stops the stack via the wrapper scripts
+# after reboot (and when docker.service is restarted). Off by default.
+install_scylla_monitoring_service: false
```
ManjotS, line 32: ❓ Question — if someone enables this and later flips it back to `false`, nothing disables or removes the previously-installed `/etc/systemd/system/scylla-monitoring.service` unit — it stays enabled and will still fire on the next reboot. Is that intentional (opt-in only, no un-opt-in path), or should there be a cleanup task guarded by `when: not install_scylla_monitoring_service | bool`?

ansible-scylla-monitoring/templates/scylla-monitoring.service.j2
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Scylla Monitoring Stack
+Requires=docker.service
+PartOf=docker.service
```
ManjotS, line 4: 💡 Praise — `Requires=`/`After=`/`PartOf=docker.service` is the right combination here: it gets start ordering, a hard start dependency, and — importantly — stop/restart propagation from `docker.service`, which is exactly what's needed for the `systemctl restart docker` scenario in the test plan to actually work.

ansible-scylla-monitoring/tasks/docker.yml
ManjotS: Applied `state: started` in a547c62. Confirmed safe to run twice in one play: `start-all.sh` runs `docker compose up -d`, which is idempotent (no-op if containers already up). Kept the manual `command` start too so behavior is unchanged when the flag is off.

ansible-scylla-monitoring/tasks/docker.yml
ManjotS: Switched to short `systemd` module name in a547c62 to match `service:` above.

ansible-scylla-monitoring/tasks/docker.yml
ManjotS: Added coverage in a547c62: `converge.yml` now sets `install_scylla_monitoring_service: true` for the monitor host, and `verify.yml` asserts the unit file exists and is `enabled`.

ansible-scylla-monitoring/defaults/main.yml
ManjotS: Good catch — added an un-opt-in path in a547c62. When the flag is `false`, the unit is disabled (only if present) and the file removed (with a `reload systemd` handler). The running stack is left untouched since the start wrapper already (re)started it earlier in the same run.

ansible-scylla-monitoring/templates/scylla-monitoring.service.j2
ManjotS: Thanks — leaving the unit dependencies as-is.

ansible-scylla-monitoring/templates/scylla-monitoring.service.j2
```diff
@@ -0,0 +1,15 @@
...
+After=network-online.target docker.service
+Wants=network-online.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
```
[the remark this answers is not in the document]
ManjotS: Valid — fixed in 004d9ac. The unit now runs as the deploy user (`User={{ ansible_user_id }}`) in the default (non-sudo) path, so root no longer executes a user-owned script. When `run_docker_with_sudo` is set (docker needs root, unit runs as root), the start/stop wrappers are written `root:root` `0750` so the deploy user can't modify a script root executes. Both paths now match the manual start privilege.

ansible-scylla-monitoring/tasks/docker.yml
```diff
@@ -79,3 +86,42 @@
...
+    mode: "0644"
+  become: true
+  when: install_scylla_monitoring_service | bool
+
+- name: "docker.yml: Enable scylla-monitoring at boot"
+  systemd:
+    name: scylla-monitoring.service
+    enabled: true
+    state: started
+    daemon_reload: true
```
vreniers: I'm wondering why one task does `daemon_reload` and another notifies the handler to do so.

Can this be made consistent?

ansible-scylla-monitoring/tasks/docker.yml
```diff
@@ -51,7 +56,9 @@
     src: start-monitoring.sh.j2
     dest: "{{ scylla_monitoring_start_script_path }}"
     mode: '0750'
-    owner: "{{ ansible_user_id }}"
+    owner: "{{ 'root' if run_docker_with_sudo | default(false) | bool else ansible_user_id }}"
+    group: "{{ 'root' if run_docker_with_sudo | default(false) | bool else ansible_user_id }}"
+  become: "{{ run_docker_with_sudo | default(false) | bool }}"
```
vreniers: The default is already specified in the `main.yml` and it's False by default. 

I'm wondering why we re-define a default.

ansible-scylla-monitoring/tasks/docker.yml
```diff
@@ -79,3 +86,42 @@
...
+    - not install_scylla_monitoring_service | bool
+    - ansible_facts.services["scylla-monitoring.service"] is defined
+
+- name: "docker.yml: Remove scylla-monitoring systemd unit"
+  file:
+    path: /etc/systemd/system/scylla-monitoring.service
+    state: absent
+  become: true
+  when: not install_scylla_monitoring_service | bool
+  notify: reload systemd
```
vreniers: Can you group these 4 tasks into a block of 2 tasks respectively each?

One block of tasks for when we want to install the monitoring service, and one block for when we don't.

vreniers (changes requested):
Some cosmetic comments. Please also address the CICD failures in this PR.

Cursor also found the following:


## 1. Env file is still a root-LPE vector (security review missed it)
The security fix made the start/stop wrappers root:root when run_docker_with_sudo, but the env file was left owner: {{ ansible_user_id }}, mode 0600 — and the start wrapper sources it:


start-monitoring.sh.j2
Lines 5-8
```
if [ -f "$ENV_FILE" ]; then
  # shellcheck disable=SC1090
  . "$ENV_FILE"
fi
```

. executes the file as shell. With run_docker_with_sudo=true the unit (and the manual start task) runs the wrapper as root, root sources a user-owned, user-writable file → same privilege escalation the review flagged for the scripts. Fix: apply the identical owner/group = root if run_docker_with_sudo + become treatment to the start-monitoring.env task (lines 41-47 of docker.yml).

## 2. Double-start conflict when flag enabled
docker.yml unconditionally starts the stack via the command task, then the new task does state: started on the oneshot unit:


docker.yml
Lines 76-78
- name: "docker.yml: start scylla-monitoring"
  command: "{{ scylla_monitoring_start_script_path }}"
Unit was never systemd-started, so systemd sees it inactive and runs ExecStart = start wrapper = start-all.sh again on an already-running stack. start-all.sh isn't idempotent → container/port conflict. You already have follow-ups #540 (idempotent wrapper) and #541 (double-start fix) as drafts — so this is known, but as-is #536 alone can error on deploy with the flag on. Either land 536 after those, or don't state: started here (unit is already effectively running via the command task; just enabled: true).

ansible-scylla-monitoring/tasks/docker.yml
ManjotS: Valid — made consistent in 554349f. Both systemd tasks now do `daemon_reload: true` inline and the `reload systemd` handler is gone (handlers/main.yml back to its original state).

The reason it can't go the other way (handler for both) is that the install path enables a brand-new unit file: without a reload first, `systemctl show` still reports `LoadState=not-found` and the systemd module fails with "Could not find the requested service". Handlers only flush at the end of the play, so the inline reload is required there — hence using it in both places rather than mixing the two.

ansible-scylla-monitoring/tasks/docker.yml
ManjotS: Agreed, that was redundant. Dropped in 554349f — all three template tasks and the unit template now use plain `run_docker_with_sudo | bool`, relying on the `False` default in `defaults/main.yml`.

Note the pre-existing `start scylla-monitoring` task below still has its own `run_docker_with_sudo is defined and ...` guard; left that untouched since it's outside this change.

ansible-scylla-monitoring/tasks/docker.yml
ManjotS: Done in 554349f. Now two blocks:

- `Install and enable scylla-monitoring at boot` — `become: true` + `when: install_scylla_monitoring_service | bool` on the block, containing the template install and the enable task.
- `Disable and remove scylla-monitoring boot service` — `become: true` + `when: not install_scylla_monitoring_service | bool` on the block, containing the disable and the file removal.

The only remaining task-level `when` is the `ansible_facts.services[...] is defined` guard on the disable task, since `systemctl disable` needs the unit to still be known to systemd.

ManjotS:
@vreniers thanks — all four points plus the CI failure are addressed in 554349f. Inline replies cover the three cosmetic ones; the two findings from your review body:

**1. Env file was still a root-LPE vector — confirmed and fixed.**

You're right, the security review missed it. `start-monitoring.sh` does `. "$ENV_FILE"`, which executes the file as shell, and with `run_docker_with_sudo=true` that wrapper runs as root (both from the systemd unit and from the manual `command` task), while the env file was left `owner: {{ ansible_user_id }}`. Same escalation path as the wrapper scripts. The env file task now gets identical treatment:

```yaml
owner: "{{ 'root' if run_docker_with_sudo | bool else ansible_user_id }}"
group: "{{ 'root' if run_docker_with_sudo | bool else ansible_user_id }}"
become: "{{ run_docker_with_sudo | bool }}"
```

Mode stays `0600`, so the CQL credentials are still only readable by the account that runs the stack.

**2. Double-start conflict — confirmed, and it was the CI failure.**

This wasn't hypothetical. The last completed molecule run on this branch failed on exactly that, on both debian11 and ubuntu2404:

```
fatal: [scylla-monitor-node]: FAILED! => {"changed": false, "msg": "Unable to start service scylla-monitoring.service: Job for scylla-monitoring.service failed because the control process exited with error code."}
```

The stack was already up from the `command` task, systemd saw the oneshot unit as inactive, ran `ExecStart` = `start-all.sh` again, and hit the container/port conflict. Took your suggested option: dropped `state: started` and kept `enabled: true`, so this PR no longer depends on #540/#541 landing first. Trade-off noted in a comment above the block — `systemctl status scylla-monitoring` will read `inactive (dead)` right after a deploy even though the stack is running, until the first reboot or `systemctl restart docker`. Once #540 makes the wrapper idempotent, #541 can make systemd the single owner of start/stop and remove that discrepancy.

The other molecule failures in that run (`pkill: not found`, `node_exporter_install: not found`, `io.conf` grep, node1/2/3 port 10000 timeouts) are on the scylla node hosts and unrelated to this diff — they also show up on the earlier run of this branch before any of these tasks existed. A fresh run is queued on 554349f.

ManjotS:
### CI status: blocked on infrastructure, not on this PR

The molecule checks can't run right now. The repo has a single self-hosted runner and it's **offline**:

```
name: fe-gh-actions-ansible-roles-u24   status: offline   busy: false
```

That has backed up the queue since Jul 25 — 10+ molecule runs are sitting in `queued` across `dev`, this branch, and several `cursor/ci-failure-automation-*` branches, none of them ever picking up a runner. The run for 554349f (30300331529) is queued behind all of them. I cancelled this branch's superseded run on the old `004d9ac` commit to take 6 stale jobs out of the queue, but the runner itself needs to come back before anything turns green. Flagging in case someone with access can restart it.

### Local validation instead

Since CI can't confirm the fix, I reproduced the failing path locally in the same image CI uses (`jrei/systemd-debian:11`, real systemd via `--privileged --cgroupns=host`), driving the exact task YAML from 554349f. Three scenarios, all passing:

```
PLAY [Enable path: unit installed and enabled but NOT started]
  ok: unit installed, is-enabled=enabled, is-active=inactive,
      start wrapper never invoked by systemd

PLAY [Reboot simulation: unit actually starts the stack on boot]
  ok: unit ExecStart invoked the start wrapper, reboot path works

PLAY [Disable path: unit disabled and removed when flag is off]
  ok: unit file removed, is-enabled reports no such file, enable symlink gone
```

The first one is the regression that failed CI on debian11 and ubuntu2404 — the wrapper is provably never invoked during the play now, so there's no second `start-all.sh` to collide with the running stack. The second confirms dropping `state: started` didn't cost us the actual feature: systemd still runs `ExecStart` when the unit is started at boot. The third confirms the un-opt-in path, and that the running stack is deliberately left up (`is-active` stays `active` after the unit is removed), which matches the comment in the code.

Also verified along the way:

- **The inline `daemon_reload` is load-bearing.** Enabling the brand-new unit works because of it; this is the reason I couldn't move both tasks onto a handler, per the thread above.
- **`run_docker_with_sudo: true` branch** renders the unit with no `User=`/`Group=` (runs as root, as intended), and the ownership expression resolves to `root` for the wrapper and env-file tasks.

Caveat on scope: this exercises the systemd install/enable/disable/remove logic and the unit template, not a real `start-all.sh` against live containers — the wrappers were stubs that record invocation. The full docker-compose path still needs the molecule run once the runner is back.

PR #537, opened 2026-07-25
scylla-manager: automatically set Alternator credentials (#528)

igorribeiroduarte:
Cherry-pick https://github.com/scylladb/scylla-ansible-roles/pull/528 to master.

### DIFF
--- a/ansible-scylla-manager/tasks/add_cluster.yml
+++ b/ansible-scylla-manager/tasks/add_cluster.yml
@@ -8,6 +8,16 @@
     echo $(sctool status|grep -q "Cluster: {{ item.cluster_name }} " && echo "present")
   register: cluster_already_added
 
+- name: Fetch alternator salted hash for "{{ item.username }}"
+  shell: |
+    cqlsh {{ item.host }} -u '{{ item.username }}' -p '{{ item.password }}' \
+      -e "SELECT salted_hash FROM system.roles WHERE role = '{{ item.username }}';" \
+      | grep -E '^\s+\$' | tr -d ' '
+  register: alternator_salted_hash
+  when:
+    - item.username is defined
+    - item.password is defined
+
 - name: run sctool cluster add for "{{ item.cluster_name }}"
   shell: |
     sctool cluster add \
@@ -19,6 +29,12 @@
       {% if item.username is defined and item.password is defined %}
       --username {{ item.username }} \
       --password {{ item.password }} \
+      {% endif %}
+      {% if alternator_salted_hash is defined
+         and alternator_salted_hash.stdout is defined
+         and alternator_salted_hash.stdout != '' %}
+      --alternator-access-key-id {{ item.username }} \
+      --alternator-secret-access-key '{{ alternator_salted_hash.stdout }}' \
       {% endif %}
   when: cluster_already_added is not defined or cluster_already_added.stdout != "present"
 
@@ -32,6 +48,12 @@
       {% if item.username is defined and item.password is defined %}
       --username {{ item.username }} \
       --password {{ item.password }} \
+      {% endif %}
+      {% if alternator_salted_hash is defined
+         and alternator_salted_hash.stdout is defined
+         and alternator_salted_hash.stdout != '' %}
+      --alternator-access-key-id {{ item.username }} \
+      --alternator-secret-access-key '{{ alternator_salted_hash.stdout }}' \
       {% endif %}
   when: cluster_already_added is defined and cluster_already_added.stdout == "present"
       
--- a/ansible-scylla-manager/tasks/add_cluster.yml
+++ b/ansible-scylla-manager/tasks/add_cluster.yml
@@ -11,9 +11,15 @@
 - name: Fetch alternator salted hash for "{{ item.username }}"
   shell: |
     cqlsh {{ item.host }} -u '{{ item.username }}' -p '{{ item.password }}' \
+      {% if item.ssl is defined and item.ssl|bool %}--ssl{% endif %} \
       -e "SELECT salted_hash FROM system.roles WHERE role = '{{ item.username }}';" \
       | grep -E '^\s+\$' | tr -d ' '
   register: alternator_salted_hash
+  # Alternator creds are best-effort: TLS-only clusters, transient network/auth
+  # failures, or missing SELECT on system.roles must not abort the cluster add/update.
+  failed_when: false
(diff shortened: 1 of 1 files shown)


ansible-scylla-manager/tasks/add_cluster.yml
```diff
@@ -8,6 +8,16 @@
     echo $(sctool status|grep -q "Cluster: {{ item.cluster_name }} " && echo "present")
   register: cluster_already_added
 
+- name: Fetch alternator salted hash for "{{ item.username }}"
+  shell: |
+    cqlsh {{ item.host }} -u '{{ item.username }}' -p '{{ item.password }}' \
+      -e "SELECT salted_hash FROM system.roles WHERE role = '{{ item.username }}';" \
+      | grep -E '^\s+\$' | tr -d ' '
+  register: alternator_salted_hash
```
ManjotS, line 17: 🟠 **Important** — The fetched Alternator secret (`alternator_salted_hash.stdout`) isn't protected with `no_log: true`, and it's then interpolated unmasked into the `sctool cluster add`/`cluster update` command lines below (`--alternator-secret-access-key '...'`). With Ansible's default logging, a `-v`/`-vvv` run, or any log-aggregation/callback plugin capturing task output, this secret — which grants Alternator (DynamoDB-API) access to the cluster — ends up in plaintext logs. It also becomes visible in the process listing (`ps aux`) on the host running the task while `sctool` executes.

This mirrors the pre-existing gap for `--username`/`--password` a few lines below (also unprotected), so it's not a regression this PR invented from scratch — but this PR does add a second secret into the same weak pattern. Worth adding `no_log: true` to this task and the two `sctool` tasks, ideally in this PR or as an immediate follow-up.

```suggestion
  register: alternator_salted_hash
  no_log: true
```

ansible-scylla-manager/tasks/add_cluster.yml
```diff
@@ -8,6 +8,16 @@
     echo $(sctool status|grep -q "Cluster: {{ item.cluster_name }} " && echo "present")
   register: cluster_already_added
 
+- name: Fetch alternator salted hash for "{{ item.username }}"
+  shell: |
+    cqlsh {{ item.host }} -u '{{ item.username }}' -p '{{ item.password }}' \
```
ManjotS, line 13: 🟠 **Important** — This lookup has no `failed_when`/`ignore_errors`, and by default a `shell` task fails the play on non-zero exit. Two ways that bites here:

- **No TLS support**: `cqlsh` is invoked with no `--ssl` option. Any cluster with client-to-node encryption enabled will refuse a plaintext CQL connection, so this task fails outright for those clusters.
- **Any other transient/auth failure** (network blip to `item.host:9042`, the CQL user lacking `SELECT` on `system.roles`, wrong password) now aborts the whole "add cluster"/"enforce configuration" flow for that item — even though, before this PR, a cluster with just `username`/`password` and no Alternator involvement would have added/updated successfully every time.

Since Alternator credentials are meant to be best-effort/optional (mirroring the `alternator_salted_hash is defined` guard used later), consider `ignore_errors: true` (or `failed_when: false`) here so a failed lookup just skips the `--alternator-*` flags instead of failing the run — and/or documenting the new `cqlsh`/TLS/permission prerequisites this task now imposes on every username+password cluster.

ManjotS (commented):
Cherry-pick of #528 (already reviewed/approved there). I confirmed this brings over exactly the `add_cluster.yml` hunk — the other two files #528 touched (`ansible-scylla-node/tasks/common.yml`, `start_one_node.yml`) aren't included here, but that's expected: #524 already landed those fixes on `master` separately, so there's nothing missing.

The new logic (query the CQL role's salted hash via `cqlsh` and feed it to `sctool` as the Alternator secret key) is sound and correctly gated on `item.username`/`item.password` being set. Two things worth addressing before/soon after merge, both centered on the new "Fetch alternator salted hash" task:

1. The freshly-fetched Alternator secret isn't shielded from Ansible's own output (verbose logs, `ansible-playbook -v`, any log/callback aggregation), and it also lands unmasked in the `sctool` command lines that consume it.
2. The `cqlsh` lookup has no failure handling or TLS support, so on clusters with client-encryption enabled, or any transient auth/network hiccup, it now hard-fails the whole per-cluster add/update instead of degrading gracefully like the optional flags around it.

Reviewed: correctness/logic, security, reliability, maintainability, testing (none added, consistent with the rest of this file). Scope was small enough that performance/docs weren't relevant.

ansible-scylla-manager/tasks/add_cluster.yml
```diff
@@ -20,6 +30,12 @@
       --username {{ item.username }} \
       --password {{ item.password }} \
       {% endif %}
+      {% if alternator_salted_hash is defined
+         and alternator_salted_hash.stdout is defined
+         and alternator_salted_hash.stdout != '' %}
+      --alternator-access-key-id {{ item.username }} \
+      --alternator-secret-access-key '{{ alternator_salted_hash.stdout }}' \
```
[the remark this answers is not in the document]
ManjotS: Fixed in 9cdbb7d. Added `no_log: true` to the salted-hash fetch task and both `sctool cluster add`/`cluster update` tasks, so `alternator_salted_hash.stdout` and the `--alternator-secret-access-key` command line no longer reach Ansible logs, `-vvv` output, or callback plugins.

ansible-scylla-manager/tasks/add_cluster.yml
ManjotS: Done in 9cdbb7d — applied `no_log: true` to the fetch task and both `sctool` add/update tasks (covers the alternator secret plus the pre-existing `--username`/`--password` exposure in those command lines).

ansible-scylla-manager/tasks/add_cluster.yml
ManjotS: Fixed in 9cdbb7d. Added `failed_when: false` to the lookup so a failed salted-hash fetch just skips the optional `--alternator-*` flags (mirroring the `alternator_salted_hash ... != ''` guard) instead of aborting the add/update. Also added optional TLS support via `item.ssl` (`--ssl`) for client-encryption-enabled clusters; other transient/auth/permission failures now degrade gracefully too.

vreniers:
@ManjotS this is a cherry-picked commit that already underwent review. 

I would have personally tested the entire `dev` branch and then merged it into `master`.

vreniers:
I merged this via another PR https://github.com/scylladb/scylla-ansible-roles/pull/545

PR #538, opened 2026-07-25
fix(replace): own service startup suppression

ManjotS:
## Summary
- add a role-level `skip_start_scylla_service` safety gate
- make `replace_node.yml` set the gate while installing/configuring the replacement node
- remove the requirement for reused deployment params to set `start_scylla_service: false`

## Context
Reimplements scylladb/field-engineering#2940 after scylladb/field-engineering#3218 was reverted.

The earlier change relied on `example-playbooks/replace_node/vars/main.yml`, but automation passes `nodes_params.yml` with `-e @file`. Those extra vars have higher precedence and commonly set `start_scylla_service: true`, causing Scylla to start before replacement metadata is configured.

This change gives the replace playbook an independent role control. It suppresses automatic Scylla startup, related replication adjustment, and Manager agent startup even when reused deployment parameters enable normal service startup. The playbook still starts the replacement explicitly after setting `replace_node_first_boot` or `replace_address_first_boot`.

## Test plan
- [x] Parse all modified YAML files with PyYAML
- [x] Run `ansible-playbook --syntax-check` with repository async plugins and required `new_node` variable

### DIFF
--- a/ansible-scylla-node/defaults/main.yml
+++ b/ansible-scylla-node/defaults/main.yml
@@ -167,6 +167,10 @@ cleanup_timeout_seconds: 36000
 # this should be set to false
 start_scylla_service: true
 
+# Let operation playbooks suppress automatic service startup even when reused
+# deployment parameters set start_scylla_service to true.
+skip_start_scylla_service: false
+
 # Only relevant for Debian/Ubuntu
 scylla_repo_keyserver: 'hkp://keyserver.ubuntu.com:80'
 scylla_repo_keys:
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -426,7 +426,7 @@
     _keyspace_replication_strategy: "{{ system_auth_replication_strategy }}"
     _keyspace_rf: "{{ system_auth_rf }}"
   when:
-    - start_scylla_service|bool
+    - start_scylla_service is defined and start_scylla_service|bool
     - not skip_start_scylla_service|bool
     - adjust_system_auth_replication is defined and adjust_system_auth_replication|bool
     - _authentication_enabled is defined and _authentication_enabled|bool
@@ -439,7 +439,7 @@
     _keyspace_replication_strategy: "{{ audit_replication_strategy }}"
     _keyspace_rf: "{{ audit_rf }}"
   when:
-    - start_scylla_service|bool
+    - start_scylla_service is defined and start_scylla_service|bool
     - not skip_start_scylla_service|bool
     - adjust_audit_replication is defined and adjust_audit_replication|bool
     - _audit_enabled is defined and _audit_enabled|bool
@@ -452,7 +452,7 @@
     _keyspace_replication_strategy: "{{ system_distributed_replication_strategy }}"
     _keyspace_rf: "{{ system_distributed_rf }}"
   when:
-    - start_scylla_service|bool
+    - start_scylla_service is defined and start_scylla_service|bool
     - not skip_start_scylla_service|bool
     - adjust_system_distributed_replication is defined and adjust_system_distributed_replication|bool
     - system_distributed_rf is defined and system_distributed_replication_strategy is defined
@@ -464,7 +464,7 @@
     _keyspace_replication_strategy: "{{ system_traces_replication_strategy }}"
     _keyspace_rf: "{{ system_traces_rf }}"
   when:
-    - start_scylla_service|bool
+    - start_scylla_service is defined and start_scylla_service|bool
     - not skip_start_scylla_service|bool
     - adjust_system_traces_replication is defined and adjust_system_traces_replication|bool
     - system_traces_rf is defined and system_traces_replication_strategy is defined
--- a/ansible-scylla-node/tasks/common.yml
+++ b/ansible-scylla-node/tasks/common.yml
@@ -395,7 +395,9 @@
         line: ""
       become: true
       when: token_distributor is defined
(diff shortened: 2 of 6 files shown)


ansible-scylla-node/defaults/main.yml
```diff
@@ -167,6 +167,10 @@ cleanup_timeout_seconds: 36000
 # this should be set to false
 start_scylla_service: true
 
+# Let operation playbooks suppress automatic service startup even when reused
+# deployment parameters set start_scylla_service to true.
+skip_start_scylla_service: false
```
ManjotS, line 172: 💡 **Praise** — Naming this `skip_start_scylla_service` (rather than reusing/overloading `start_scylla_service`) is the right call: it gives operation playbooks an independent, additive gate instead of fighting over precedence of the same variable, which is exactly what caused the regression this PR fixes.

ansible-scylla-node/tasks/common.yml
```diff
@@ -425,6 +427,7 @@
     _keyspace_rf: "{{ system_auth_rf }}"
   when:
     - start_scylla_service|bool
```
ManjotS: 🔵 nit (pre-existing, out of scope for this PR) — this condition (`start_scylla_service|bool`, and the same pattern on lines 442, 455, 467) omits the `is defined` guard that's used at line 399 and in `manager_agents.yml` (`start_scylla_service is defined and start_scylla_service|bool`). Since `start_scylla_service` has a role default, this doesn't currently bite, but it's an inconsistency worth normalizing at some point for defense-in-depth if the default is ever removed.

ManjotS (commented):
Reviewed the role/playbook changes for the `skip_start_scylla_service` gate. This is a clean, well-scoped fix: it introduces an independent role-level variable that suppresses automatic Scylla startup, replication adjustment, and Manager-agent startup, and threads it through every `when:` clause that previously gated on `start_scylla_service` alone (the big `start_scylla_service dependent tasks` block, the four `Adjust replication for *` includes, and the Manager agent start task in `manager_agents.yml`).

Why this works: `skip_start_scylla_service` is set via role `vars:` on the role invocation in `replace_node.yml` (`roles: - role: ansible-scylla-node vars: {skip_start_scylla_service: true}`). In Ansible's variable precedence, extra-vars (`-e @nodes_params.yml`) always win over role params — which is exactly the bug being fixed (reused `start_scylla_service: true` from `nodes_params.yml` was clobbering the playbook's own `false`). Since `skip_start_scylla_service` is a brand-new variable name that existing `nodes_params.yml` files won't define, it can't be shadowed the same way. Good fix for the root cause rather than a workaround.

Verification performed:
- Confirmed `ansible-playbook --syntax-check` passes on `replace_node.yml` with the repo's async action plugins loaded.
- Traced every `start_scylla_service` reference in the role to confirm the new gate is applied everywhere it needs to be, and that `start_one_node.yml` (used to explicitly start the new node later, and to start it during the final wait/repair play) is unaffected since it's invoked directly via `tasks_from`, not through the gated block.
- Confirmed the second play in `replace_node.yml` (`hosts: scylla`) applying the gate to all hosts, not just `new_node`, matches the pre-existing behavior of the old `vars/main.yml` `start_scylla_service: false` (which also applied playbook-wide), so this isn't a behavior change in scope.
- Confirmed `skip_start_scylla_service: false` as a new role default doesn't affect any other playbook that consumes this role.

Coverage: correctness/logic, security, performance, maintainability, and docs (README) all reviewed — no concerns. No new tests were added, but this is a variable-precedence/control-flow fix in Ansible YAML where the existing syntax-check + manual training-cluster run (noted as pending in the test plan) is the appropriate validation; I don't think unit tests are expected here. No blocking or important issues found.

ansible-scylla-node/tasks/common.yml
ManjotS: Normalized in e108729: added `start_scylla_service is defined and` to the four replication-adjust when clauses (lines 429/442/455/467), matching line 399 and manager_agents.yml.

vreniers (changes requested):
I don't think this is the right approach to the problem. This PR introduces yet another control variable, to skip another control variable. That's a quick path to make things confusing.

Instead, we should not persist control variables as part of a state file (i.e. node_params.yml). When adding nodes, one should pass this rather explicitly for example. Not on each repetitive action. 

Upon closer look, if we stop using `-e` and adopt groupvars, the problem also would be solved here. I think this is a better strategy even. 

Cf. https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence

ManjotS:
Agree with this. `skip_start_scylla_service` is a workaround for the `-e` precedence problem, not a fix — it stacks a control var on top of a control var, which is exactly the confusion you're calling out.

The right fix is what you said: stop persisting `start_scylla_service` in `nodes_params.yml` and let groupvars carry it once we're off `-e` for node params. Precedence then works in our favor instead of against us.

The reason I went the skip-var route: prod clusters are still on the `-e` path, and FE force-passes it — `fc_playbook_executor.py` sets `start_scylla_service=false` for replace, and `nodes_params.yml` commonly carries `true`. So restoring `vars/main.yml: start_scylla_service: false` on its own just reproduces the scylladb/field-engineering#3218 regression the moment reused params win.

So I'm closing this. The fix belongs in the groupvars migration + dropping `start_scylla_service` from the `-e` state on the FE side, not another gate in the role. Tracking in scylladb/field-engineering#2940.

PR #542, opened 2026-07-28
ci: fix molecule DinD volume leak and broaden test-roles triggers (ANSROLES-20)

vreniers:
## Summary
- **ANSROLES-20:** Use per-scenario named DinD volumes (`molecule-dind-<distro>:/var/lib/docker`) on `scylla-monitor-node` and remove them in an `if: always()` step after molecule, so orphan volumes stop filling the self-hosted runner disk.
- **Triggers (separate commit):** Run molecule when `molecule/**` or `test-roles.yml` change; allow `pull_request` targeting `dev`; keep the workflow path filter scoped to this file (not all of `.github/workflows/`).

## Commits
1. `ci: clean up molecule DinD volumes after tests (ANSROLES-20)`
2. `ci: run molecule on harness changes and PRs to dev`

## Test plan
- [ ] Molecule jobs run on this PR (base `dev` + path filters)
- [ ] After each matrix job, runner has no leftover `molecule-dind-<distro>` volume
- [ ] Monitoring DinD still works (volume still present for overlay; only name/cleanup changed)
- [ ] Editing an unrelated workflow does **not** start this matrix

### DIFF
--- a/.github/workflows/test-roles.yml
+++ b/.github/workflows/test-roles.yml
@@ -46,3 +46,8 @@ jobs:
       - name: Test with molecule
         run: |
           molecule destroy --scenario-name ${{ matrix.molecule_distro }} && molecule test --scenario-name ${{ matrix.molecule_distro }}
+      # ANSROLES-20: molecule destroy drops scylla-monitor-node but keeps its named
+      # DinD volume (molecule-dind-<distro>). Remove explicitly so runner disk does not fill.
+      - name: Remove molecule DinD volume
+        if: always()
+        run: docker volume rm -f "molecule-dind-${{ matrix.molecule_distro }}" || true
--- a/.github/workflows/test-roles.yml
+++ b/.github/workflows/test-roles.yml
@@ -6,14 +6,18 @@ on:
       - 'ansible-scylla-node/**'
       - 'ansible-scylla-monitoring/**'
       - 'ansible-scylla-manager/**'
-      - '.github/workflows/*'
+      - 'molecule/**'
+      - '.github/workflows/test-roles.yml'
   pull_request:
     paths:
       - 'ansible-scylla-node/**'
       - 'ansible-scylla-monitoring/**'
       - 'ansible-scylla-manager/**'
+      - 'molecule/**'
+      - '.github/workflows/test-roles.yml'
     branches:
       - master
+      - dev
 jobs:
   build:
     runs-on: self-hosted
--- a/.github/workflows/test-roles.yml
+++ b/.github/workflows/test-roles.yml
@@ -51,7 +51,10 @@ jobs:
         run: |
           molecule destroy --scenario-name ${{ matrix.molecule_distro }} && molecule test --scenario-name ${{ matrix.molecule_distro }}
       # ANSROLES-20: molecule destroy drops scylla-monitor-node but keeps its named
-      # DinD volume (molecule-dind-<distro>). Remove explicitly so runner disk does not fill.
+      # DinD volume (molecule-dind-<distro>-<run_id>). Remove explicitly so runner disk
+      # does not fill. Warn (do not swallow) if removal fails for a real reason.
       - name: Remove molecule DinD volume
         if: always()
-        run: docker volume rm -f "molecule-dind-${{ matrix.molecule_distro }}" || true
+        run: |
+          volume="molecule-dind-${{ matrix.molecule_distro }}-${{ github.run_id }}"
+          docker volume rm -f "$volume" || echo "::warning::failed to remove volume $volume"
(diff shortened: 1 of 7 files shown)


.github/workflows/test-roles.yml
```diff
@@ -6,14 +6,18 @@ on:
...
       - 'ansible-scylla-monitoring/**'
       - 'ansible-scylla-manager/**'
+      - 'molecule/**'
+      - '.github/workflows/test-roles.yml'
     branches:
       - master
+      - dev
 jobs:
   build:
     runs-on: self-hosted
```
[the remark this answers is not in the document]
vreniers: This pattern already pre-existed. The runner has limited permissions.

.github/workflows/test-roles.yml
```diff
@@ -46,3 +50,8 @@ jobs:
       - name: Test with molecule
         run: |
           molecule destroy --scenario-name ${{ matrix.molecule_distro }} && molecule test --scenario-name ${{ matrix.molecule_distro }}
+      # ANSROLES-20: molecule destroy drops scylla-monitor-node but keeps its named
+      # DinD volume (molecule-dind-<distro>). Remove explicitly so runner disk does not fill.
+      - name: Remove molecule DinD volume
+        if: always()
+        run: docker volume rm -f "molecule-dind-${{ matrix.molecule_distro }}" || true
```
ManjotS: 🟡 **Minor** — `|| true` silently swallows *any* failure of `docker volume rm -f`, not just the "volume already gone" case. If removal ever fails for a real reason (e.g. the volume is still attached because `molecule destroy` didn't fully tear down the container), there's no signal in the logs — the leak this step exists to fix could quietly come back and nobody would notice until disk fills up again.

Consider surfacing a warning instead of swallowing silently:
```suggestion
        run: docker volume rm -f "molecule-dind-${{ matrix.molecule_distro }}" || echo "::warning::failed to remove volume molecule-dind-${{ matrix.molecule_distro }}"
```

molecule/debian11/molecule.yml
```diff
@@ -56,7 +56,10 @@ platforms:
     volumes:
       - /sys/fs/cgroup:/sys/fs/cgroup
       - /sys/kernel/security:/sys/kernel/security:ro # Prevent container from modifying host AppArmor profile
-      - /var/lib/docker # https://github.com/docker/for-linux/issues/1138
+      # DinD: nested dockerd needs /var/lib/docker on a volume (overlay cannot sit on
+      # the container rootfs). See https://github.com/docker/for-linux/issues/1138
+      # Named (not anonymous) so CI can docker volume rm after destroy — ANSROLES-20.
+      - molecule-dind-debian11:/var/lib/docker
```
ManjotS: 🔵 **Nit** (applies to all six `molecule/*/molecule.yml` files, flagging once here) — moving from an anonymous to a deterministic per-distro volume name fixes the leak, but it also means two concurrent jobs for the same distro (e.g. two workflow runs overlapping on the same Docker host) would now contend for the *same* volume name, where previously each got its own anonymous volume. The cleanup step's `docker volume rm -f` from one run could then race with another run's container still using that volume.

Not an active risk today — I checked and this repo currently has only one self-hosted runner (`fe-gh-actions-ansible-roles-u24`) registered, so jobs serialize and there's no real overlap. But it's worth keeping in mind if the runner pool ever grows (the existing `max-parallel: 4` in the workflow suggests that's the eventual intent). If/when that happens, appending `${{ github.run_id }}` to the volume name would keep it unique-but-named (still cleanable) without reintroducing the collision risk. No action needed now — just flagging for future-proofing.

ManjotS (approved):
Small, well-scoped CI-only change: switches the `scylla-monitor-node` DinD `/var/lib/docker` volume from anonymous to a per-distro named volume and adds an `if: always()` step to remove it after `molecule test`, fixing the orphan-volume disk leak on the self-hosted runner (ANSROLES-20). The second commit tightens the `test-roles.yml` trigger paths (scoped to `molecule/**` and the workflow file itself, instead of all of `.github/workflows/*`) and adds `dev` to the `pull_request` `branches` filter, which also happens to be required for this very PR's CI to run against its `dev` base. Both commits do what the description says and I don't see any correctness or security problems.

Reviewed: correctness/logic, security, performance, reliability/operability, maintainability, testing (N/A — CI config), docs/style. I verified via the GitHub API that only one self-hosted runner (`fe-gh-actions-ansible-roles-u24`) is currently registered for this repo, which is relevant to one of the notes below.

Two minor/non-blocking notes inline — nothing here should hold up merging.

.github/workflows/test-roles.yml
vreniers: The cleanup step now emits a GitHub Actions ::warning:: when docker volume rm fails, so a stuck/attached volume shows up in the logs instead of silently leaking again.

molecule/debian11/molecule.yml
vreniers: Appended the run id to the DinD volume across all 6 molecule scenarios — molecule-dind-<distro>-${GITHUB_RUN_ID:-local} — and matched it in the workflow cleanup name. Overlapping same-distro runs can no longer contend for one volume name; it defaults to local outside CI. I verified molecule 4.0.4's interpolation supports the ${VAR:-default} syntax by testing against its actual source.

PR #544, opened 2026-07-28
ci: scope molecule push trigger to master and dev

vreniers:
## Summary
- Scope the `push` trigger in `test-roles.yml` to `master`/`dev` instead of removing it.
- Same-repo PR branch updates no longer double-run the molecule matrix (`push` + `pull_request`).
- Merges into `master`/`dev` still get a post-merge molecule run via scoped `push`.

## Test plan
- [ ] Push a commit to this PR → exactly one Molecule workflow run (`pull_request`)
- [ ] No `push` event run for this feature branch
- [ ] After merge to `dev`, one Molecule run with event `push` on `dev`

### DIFF
--- a/.github/workflows/test-roles.yml
+++ b/.github/workflows/test-roles.yml
@@ -2,6 +2,9 @@
 name: Molecule Test for ansible roles
 on:
   push:
+    branches:
+      - master
+      - dev
     paths:
       - 'ansible-scylla-node/**'
       - 'ansible-scylla-monitoring/**'


.github/workflows/test-roles.yml
```diff
@@ -1,13 +1,6 @@
...
 name: Molecule Test for ansible roles
 on:
-  push:
-    paths:
-      - 'ansible-scylla-node/**'
-      - 'ansible-scylla-monitoring/**'
-      - 'ansible-scylla-manager/**'
-      - 'molecule/**'
-      - '.github/workflows/test-roles.yml'
   pull_request:
```
ManjotS, line 14: 🟠 **Important** — Removing the `push` trigger entirely (rather than scoping it) also removes the Molecule run that currently fires when a PR is squash/merge/rebase-merged into `master` or `dev` — merging via any of those methods still generates a `push` event to the target branch, which is exactly what this trigger (with these same paths) was catching.

I checked this repo's branch rules: `master`'s ruleset requires a PR but has no required status check, and `dev` has no branch protection/ruleset at all. So this workflow was likely the only automated verification that actually runs against the merged state of `master`/`dev` (as opposed to the pre-merge PR branch, which can differ — conflict resolution, squash edits, or another PR landing in between).

Scoping the trigger instead of dropping it fixes the duplicate-run problem just as well, since PR source branches aren't normally named `master`/`dev`, while preserving the post-merge run:

```suggestion
  push:
    branches:
      - master
      - dev
    paths:
      - 'ansible-scylla-node/**'
      - 'ansible-scylla-monitoring/**'
      - 'ansible-scylla-manager/**'
      - 'molecule/**'
      - '.github/workflows/test-roles.yml'
  pull_request:
```

ManjotS (commented):
Small, focused CI trigger change — removes the unscoped `push:` trigger from `test-roles.yml`, leaving only `pull_request` (scoped to `master`/`dev`) to drive the Molecule matrix. The stated goal (stop double-running Molecule when pushing to a same-repo PR branch) is real and the diagnosis is correct: an unscoped `push` trigger plus `pull_request` both fire on same-repo branch pushes.

One concern worth discussing before merge (see inline comment): dropping the `push` trigger entirely, rather than scoping it to `branches: [master, dev]`, also removes the run that currently fires when a PR is merged into `master`/`dev` (squash/merge/rebase-merge all generate a `push` event to the target branch). I checked this repo's branch rules — `master` has a ruleset requiring PRs but no required status check, and `dev` has no ruleset/protection at all — so this Molecule matrix looks like the only automated signal on the actual state of those branches post-merge. A branch-scoped `push` trigger would fix the duplicate-run problem just as well (PR source branches aren't usually named `master`/`dev`) while keeping that post-merge coverage.

Coverage: reviewed correctness/CI-logic, security (n/a for this diff), reliability/CI-signal implications, and consistency with the existing fork-PR behavior described in the PR body. No automated tests apply to a workflow-trigger-only change beyond the manual test plan already listed in the description.

vreniers:
Addressed: scoped `push` to `master`/`dev` instead of dropping it, so PR branch updates stay single-run while post-merge coverage remains.

PR #531, opened 2026-07-17
Persist iptables DROP rules via iptables_drop_sources

ManjotS:
## Summary

Makes iptables DROP rules during node replace a **declared role variable**.

## Changes

1. **`ansible-scylla-common`**: new `iptables_drop_sources: []` default. Renamed `disable_firewall.yml` → `handle_firewall.yml`: when `disable_firewall` is true, flush/stop firewall services; when false, apply `INPUT`/`DROP` for each listed source via the Ansible `iptables` module (`state=present`).
2. **`replace_node.yml`**: remove the ad-hoc iptables DROP task and `block_replaced_node`. Install/Update play sets `disable_firewall: false` so DROP sources from cluster vars take effect.
3. **`ansible-scylla-node`**: no longer owns DROP logic (still exposes deprecated `firewall_enabled`, which maps to `disable_firewall` via common defaults).

## Test plan

- [ ] Molecule / lint as usual for ansible-scylla-common / ansible-scylla-node
- [ ] Manual: set `disable_firewall: false` + `iptables_drop_sources: ["10.0.0.1"]`, run role, confirm DROP present
- [ ] Manual: clear list, run role with `disable_firewall: true`, confirm flush

PR #547, opened 2026-08-12
ansible-scylla-common: remove stale iptables_drop_sources rules

ManjotS:
## Summary
Follow-up to #531, per review feedback on scylladb/field-engineering#3209: when `disable_firewall` is `false`, the role only ever added INPUT DROP rules for sources in `iptables_drop_sources` — it never removed a DROP rule for a source that was in a *previous* run's list but isn't in the current one. That left temporary blocks (e.g. field-engineering's replace-node flow) stuck in place instead of being cleared on the next run with the default empty list.

- `ansible-scylla-common/tasks/handle_firewall.yml`: reads existing simple `-A INPUT -s <ip> -j DROP` rules via `iptables -S INPUT`, diffs them against `iptables_drop_sources`, and removes any rule whose source fell off the list.
- `ansible-scylla-common/README.md`: documents that each run reconciles the INPUT chain to match `iptables_drop_sources` exactly.

## Test plan
- [x] `ansible-playbook --syntax-check` against a playbook including this role
- [x] Verified the regex/`difference` reconciliation logic in isolation with a throwaway playbook (parses a mixed set of DROP/ACCEPT rules, confirms only the stale source is selected for removal)
- [ ] Molecule run (CI)

### DIFF
--- a/ansible-scylla-common/tasks/handle_firewall.yml
+++ b/ansible-scylla-common/tasks/handle_firewall.yml
@@ -33,3 +33,31 @@
   when:
     - not (disable_firewall | bool)
     - iptables_drop_sources | length > 0
+
+# Remove any previously-applied INPUT DROP rule whose source is no longer listed in
+# iptables_drop_sources, so a source dropped by a prior run (e.g. a replace-node
+# temporary block) doesn't linger once the role's default (empty list) applies again.
+- name: Find existing simple INPUT DROP rules
+  command: iptables -S INPUT
+  register: _current_input_rules
+  changed_when: false
+  become: true
+  when: not (disable_firewall | bool)
+
+- name: Remove stale INPUT DROP rules not in iptables_drop_sources
+  iptables:
+    chain: INPUT
+    source: "{{ item }}"
+    jump: DROP
+    state: absent
+  loop: "{{ _current_drop_sources | difference(iptables_drop_sources) }}"
+  become: true
+  vars:
+    _current_drop_sources: >-
+      {{
+        _current_input_rules.stdout_lines
+        | select('match', '^-A INPUT -s [0-9.]+(/[0-9]+)? -j DROP$')
+        | map('regex_replace', '^-A INPUT -s ([0-9.]+)(?:/[0-9]+)? -j DROP$', '\1')
+        | list
+      }}
+  when: not (disable_firewall | bool)
--- a/ansible-scylla-common/tasks/handle_firewall.yml
+++ b/ansible-scylla-common/tasks/handle_firewall.yml
@@ -41,6 +41,7 @@
   command: iptables -S INPUT
   register: _current_input_rules
   changed_when: false
+  check_mode: false
   become: true
   when: not (disable_firewall | bool)
 
--- a/ansible-scylla-common/README.md
+++ b/ansible-scylla-common/README.md
@@ -11,6 +11,8 @@ A common role providing shared tasks and utilities for all Scylla Ansible roles.
 
 - `disable_firewall`: Whether firewall should be disabled. (default: true, unless deprecated `firewall_enabled` is true)
 - `iptables_drop_sources`: List of source IPs to DROP on INPUT when `disable_firewall` is false. (default: `[]`)
+  Each run reconciles the INPUT chain to match this list exactly: sources present are
+  added, and any previously-applied DROP source no longer in the list is removed.
 
 ## Usage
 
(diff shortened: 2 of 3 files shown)


ansible-scylla-common/tasks/handle_firewall.yml
```diff
@@ -33,3 +33,31 @@
   when:
     - not (disable_firewall | bool)
     - iptables_drop_sources | length > 0
+
+# Remove any previously-applied INPUT DROP rule whose source is no longer listed in
+# iptables_drop_sources, so a source dropped by a prior run (e.g. a replace-node
+# temporary block) doesn't linger once the role's default (empty list) applies again.
+- name: Find existing simple INPUT DROP rules
+  command: iptables -S INPUT
```
ManjotS, line 41: 🟠 **Important** — `iptables -S INPUT` runs through the `command` module, which does not execute under `ansible-playbook --check` (it gets skipped instead). In check mode `_current_input_rules` never gets a `stdout_lines` key, so the next task's `vars` expression (`_current_input_rules.stdout_lines | select(...)`) fails with an "has no attribute 'stdout_lines'" error, breaking any dry-run of this role. Force the read to still happen in check mode:

```suggestion
- name: Find existing simple INPUT DROP rules
  command: iptables -S INPUT
  register: _current_input_rules
  changed_when: false
  check_mode: false
  become: true
  when: not (disable_firewall | bool)
```

ansible-scylla-common/tasks/handle_firewall.yml
```diff
@@ -33,3 +33,31 @@
...
+    jump: DROP
+    state: absent
+  loop: "{{ _current_drop_sources | difference(iptables_drop_sources) }}"
+  become: true
+  vars:
+    _current_drop_sources: >-
+      {{
+        _current_input_rules.stdout_lines
+        | select('match', '^-A INPUT -s [0-9.]+(/[0-9]+)? -j DROP$')
+        | map('regex_replace', '^-A INPUT -s ([0-9.]+)(?:/[0-9]+)? -j DROP$', '\1')
```
ManjotS, line 61: 🟡 **Minor** — The regex matches an optional `/mask` suffix but the capture group (`([0-9.]+)`) only grabs the bare IP, so `regex_replace` always drops the mask. That's harmless for the documented case (bare host IPs — iptables normalizes those to `/32` in `-S` output, and stripping it round-trips against a maskless entry in `iptables_drop_sources`). But if an entry is ever given with an explicit mask (e.g. `"1.2.3.4/32"`) or is a real subnet (e.g. `"10.0.0.0/24"`), the extracted value can never equal the configured string:
- an explicit `/32` entry looks stale every run and gets removed + immediately re-added (a brief unblock window each play), and
- a genuine subnet rule is never recognized as "configured", so it can never be reconciled away either — quietly defeating this PR's own goal for that case.

Worth either a one-line note in `defaults/main.yml` restricting `iptables_drop_sources` to bare IPv4 addresses (matching current behavior), or capturing/normalizing the mask on both sides of the comparison.

ManjotS (commented):
Reviewed the stale-rule reconciliation logic added to `handle_firewall.yml`. Good fix for the leaked-block problem from #531/field-engineering#3209, and nice touch keeping the new tasks gated only on `disable_firewall` (not also on `iptables_drop_sources | length > 0`) so cleanup still runs once the list goes back to empty.

One important gap: the new read (`iptables -S INPUT` via `command`) breaks `ansible-playbook --check` runs — see inline comment. Also a minor edge case around masked/CIDR sources in the regex extraction, and a testing gap worth calling out.

Coverage: correctness/logic, security (no injection risk — command is static, `iptables` module uses structured params), performance (negligible), reliability/check-mode, maintainability, and testing. Docs change (README) looks accurate and matches the new behavior.

ansible-scylla-common/tasks/handle_firewall.yml
ManjotS: Fixed in 15dd34d: added `check_mode: false` so this task still runs under `--check`.

ansible-scylla-common/tasks/handle_firewall.yml
ManjotS: Fixed in 15dd34d: documented in defaults/main.yml that `iptables_drop_sources` entries must be bare IPv4 addresses (no mask), matching current normalization behavior.

igorribeiroduarte (approved):

ansible-scylla-common/tasks/handle_firewall.yml
```diff
@@ -33,3 +33,32 @@
...
+        | list
+      }}
+  when: not (disable_firewall | bool)
```
vreniers, line 64: It would have perhaps been better to make a bigger block:
```
- block:
  3 tasks
  when not (disable_firewall)
```

